PR #5, opened 2020-05-12
Add extended sequential and repeat

belldandyxtq:
This PR add repeat functionality to torch.nn.sequential.
The repeat behaves the same way as the one in chainer.

The tests cases are copied from chainer and modified to pytorch.

### DIFF
--- a/pytorch_pfn_extras/nn/modules/extended_sequential.py
+++ b/pytorch_pfn_extras/nn/modules/extended_sequential.py
@@ -1,6 +1,5 @@
 import torch
 import copy
-from collections.abc import Iterable
 
 
 def _reset_parameters(model):
--- a/pytorch_pfn_extras/nn/modules/extended_sequential.py
+++ b/pytorch_pfn_extras/nn/modules/extended_sequential.py
@@ -1,5 +1,20 @@
 import torch
 import copy
+from collections.abc import Iterable
+
+
+def _reset_parameters(model):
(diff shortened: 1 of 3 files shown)


pytorch_pfn_extras/nn/modules/extended_sequential.py
```diff
@@ -0,0 +1,68 @@
...
+class ExtendedSequential(torch.nn.Sequential):
+    """Sequential module with extended features from chainer.
+
+    """
+    def _copy_model(self, mode):
+        if mode == 'init':
+            return _reset_parameters(copy.deepcopy(self))
+        elif mode == 'copy':
+            return copy.deepcopy(self)
+        else:
```
emcastillo, line 29: nit:
How about checking the values and raise an error for unknown modes?

belldandyxtq: I think the mode is checked in https://github.com/pfnet/pytorch-pfn-extras/blob/b5bc392feedd0f5c982180d67f11a7d14652f00d/pytorch_pfn_extras/nn/modules/extended_sequential.py#L60

emcastillo: Fair enough!

emcastillo:
Please, fix flake8 errors

belldandyxtq:
Sorry, I just added more coverage to the test, can you please review it again? @emcastillo

PR #6, opened 2020-05-12
Include __version__ in the root module

belltailjp:
I implemented `__version__` and use it from `setup.py` based on [how it is done in Chainer](https://github.com/chainer/chainer/blob/df53bff/setup.py#L118-L120).

```python
>>> import pytorch_pfn_extras
>>> pytorch_pfn_extras.__version__
'0.2.0'
```

I guess that when this PR becomes ready, the version should be different (`0.2.1` or so?). Once it is determined, please tell me, I'd like to update the PR.


kmaehashi:
Thanks! So far we expect to bump version number at release, so I think it's ok as is.

PR #7, opened 2020-05-12
Apply transformations to models before creating and loading the `state_dict`

emcastillo:
This PR is intended to solve cases such as saving the module object of a torch.nn.DataParallel object in the snapshot.  It also allows to apply a transformation to a model after loading the snapshot.


asi1024:
LGTM.

PR #8, opened 2020-05-12
Fix flake8 syntax warnings

emcastillo:
The new flake8 version detects now issues that weren't detected before.
Fix the code to avoid these errors.


kmaehashi:
LGTM!

PR #9, opened 2020-05-13
Support `TabularDataset`

emcastillo:
Port `TabularDataset` from chainer

Migrated all the tests to pytest, removed six

### DIFF
--- a/pytorch_pfn_extras/dataset/tabular/_transform.py
+++ b/pytorch_pfn_extras/dataset/tabular/_transform.py
@@ -55,8 +55,6 @@ class _Transform(tabular_dataset.TabularDataset):
             if (key_indices is None) or len(contained) > 0:
                 # Now look the indices of the keys we need to fetch
                 # from the original dataset to apply this transformation
-                if len(ops) == 0:
-                    ops = self._dataset.keys
                 ops_idx = _utils._as_key_indices(ops, self._dataset.keys)
                 operands.update(ops_idx)
                 transforms.append((ops_idx, t, res_idx))
--- a/pytorch_pfn_extras/dataset/tabular/_transform.py
+++ b/pytorch_pfn_extras/dataset/tabular/_transform.py
@@ -3,17 +3,18 @@ from pytorch_pfn_extras.dataset.tabular import _utils
 
 
 class _Transform(tabular_dataset.TabularDataset):
-    def __init__(self, dataset, transforms):
-
+    def __init__(self, dataset, keys, transforms):
         self._dataset = dataset
-        # keys are derived from the transform signatures
-        keys = set()
+
+        key_set = set()
         for s, _ in transforms:
-            keys.update(s[1])
+            if any([k in key_set for k in s[1]]):
+                raise ValueError('Transformations must be disjoint')
+            key_set.update(s[1])
+        if len(key_set - set(keys)) != 0:
+            raise ValueError('Transformations must produce all specified keys')
 
-        # There might be issues with keys order?
-        self._keys = tuple(sorted(list(keys)))
-
+        self._keys = keys
         self._transforms = transforms
 
     def __len__(self):
@@ -45,10 +46,17 @@ class _Transform(tabular_dataset.TabularDataset):
             # We only allow to execute transformations that will generate
             # the exact requested columns
             key_indices = list(key_indices)  # sometimes we get ranges
-            contained = all([r in key_indices for r in res_idx])
-            if (key_indices is None) or contained:
+            # We look for transformations that produces the requested keys
+            # we allow key_indices select a given key for transformations
+            # producting multiple keys since we have ensured all are disjoint
+            contained = set(res_idx).intersection(key_indices)
+            # res_idx holds the transf. indexes that belong to key_indices
+            res_idx = list(contained)
+            if (key_indices is None) or len(contained) > 0:
                 # Now look the indices of the keys we need to fetch
                 # from the original dataset to apply this transformation
+                if len(ops) == 0:
+                    ops = self._dataset.keys
                 ops_idx = _utils._as_key_indices(ops, self._dataset.keys)
                 operands.update(ops_idx)
                 transforms.append((ops_idx, t, res_idx))
@@ -57,7 +65,6 @@ class _Transform(tabular_dataset.TabularDataset):
     def get_examples(self, indices, key_indices):
         if key_indices is None:
             key_indices = range(len(self._keys))
-
         ops_idx, transforms = self._find_candidate_transforms(key_indices)
         in_examples = self._dataset.get_examples(indices, ops_idx)
         out_examples = tuple([] for _ in key_indices)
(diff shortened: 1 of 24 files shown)


emcastillo:
PTAL

Hakuyume:
We used `TabularDataset` of Chainer in our project and I found a problem of its design.

`TabularDataset.transform` takes multiple columns and return multiple columns.
For examples, `TabularDataset.transform` can convert a dataset with keys (A, B, C) into another dataset with keys (D, E).
Let assume that D can calculated only by A and B, and C is very heavy to load (e.g. an image).
In this case, we want to skip loading C if we need only D.
Currently, this is not supported. `TabularDataset` always load A, B, and, C even  if C and E are not required.
In our project, we use `slice` and `join` to solve this problem. It worked well but it was hacky and not efficient. 
I think it is good time to change the API of transform since we do not need to consider backward compatibility so strictly.

emcastillo:
All the changes to `transform` have been added. I think this is ready for review now.

pytorch_pfn_extras/dataset/tabular/_transform.py
```diff
@@ -3,18 +3,19 @@
...
-    def __init__(self, dataset, keys, transform, fetch_keys=None):
-        if not isinstance(keys, tuple):
-            keys = (keys,)
-
+    def __init__(self, dataset, keys, transforms):
         self._dataset = dataset
+
+        key_set = set()
+        for s, _ in transforms:
+            if any([k in key_set for k in s[1]]):
```
Hakuyume, line 11: We can avoid constructing list sicne `any` can take a generator.
```python
any(k in key_set for k in s[1])
```

pytorch_pfn_extras/dataset/tabular/_transform.py
```diff
@@ -3,18 +3,19 @@
...
+                raise ValueError('Transformations must be disjoint')
+            key_set.update(s[1])
+        if len(key_set - set(keys)) != 0:
+            raise ValueError('Transformations must produce all specified keys')
```
Hakuyume, line 15: The condition and message are mismatched.
The condition indicates that all produced keys by transformations are specified in keys.
If `keys = ('A', 'B')` and transformations produce `('A',)`, this check passes.

pytorch_pfn_extras/dataset/tabular/_transform.py
```diff
@@ -29,49 +30,98 @@ def mode(self):
...
+        # Assume that all the registered transformations are
+        # disjoint on the outputs
+        transforms = []
+        operands = set()
+        # Look for the transforms that produce the
+        # columns specified in the result
+        # to avoid calculating uneeded columns
+        for s, t in self._transforms:
+            ops, res = s
+            res_idx = _utils._as_key_indices(res, self._keys)
```
Hakuyume, line 45: Converting name to index in `__init__` is more efficient, since `res_idx` does not change. (maybe in another PR).

pytorch_pfn_extras/dataset/tabular/_transform.py
```diff
@@ -29,49 +30,98 @@ def mode(self):
...
+            # We only allow to execute transformations that will generate
+            # the exact requested columns
+            key_indices = list(key_indices)  # sometimes we get ranges
+            # We look for transformations that produces the requested keys
+            # we allow key_indices select a given key for transformations
+            # producting multiple keys since we have ensured all are disjoint
+            contained = set(res_idx).intersection(key_indices)
+            # res_idx holds the transf. indexes that belong to key_indices
+            res_idx = list(contained)
+            if (key_indices is None) or len(contained) > 0:
```
Hakuyume, line 55: `key_indices` never become `None` here.

pytorch_pfn_extras/dataset/tabular/_transform.py
```diff
@@ -29,49 +30,98 @@ def mode(self):
...
+        operands = set()
+        # Look for the transforms that produce the
+        # columns specified in the result
+        # to avoid calculating uneeded columns
+        for s, t in self._transforms:
+            ops, res = s
+            res_idx = _utils._as_key_indices(res, self._keys)
+            # We only allow to execute transformations that will generate
+            # the exact requested columns
+            key_indices = list(key_indices)  # sometimes we get ranges
```
Hakuyume, line 48: This conversion should be done in the beginning of this function.

pytorch_pfn_extras/dataset/tabular/_transform.py
```diff
@@ -29,49 +30,98 @@ def mode(self):
...
+            contained = set(res_idx).intersection(key_indices)
+            # res_idx holds the transf. indexes that belong to key_indices
+            res_idx = list(contained)
+            if (key_indices is None) or len(contained) > 0:
+                # Now look the indices of the keys we need to fetch
+                # from the original dataset to apply this transformation
+                ops_idx = _utils._as_key_indices(ops, self._dataset.keys)
+                operands.update(ops_idx)
+                transforms.append((ops_idx, t, res_idx))
+        return sorted(list(operands)), transforms
```
Hakuyume, line 61: Do we need to sort input keys?

emcastillo:
Thanks for the comments!, I will fix them today

pytorch_pfn_extras/dataset/tabular/_transform.py
emcastillo: I think we don't, good catch

pytorch_pfn_extras/dataset/tabular/_transform.py
emcastillo: Changed it!

pytorch_pfn_extras/dataset/tabular/_transform.py
emcastillo: done, all is pre-processed

emcastillo:
@Hakuyume comments addressed, thanks!

pytorch_pfn_extras/dataset/tabular/_transform.py
```diff
@@ -0,0 +1,220 @@
...
+        # disjoint on the outputs
+        key_indices = list(key_indices)  # sometimes we get ranges
+        transforms = []
+        operands = set()
+        # Look for the transforms that produce the
+        # columns specified in the result
+        # to avoid calculating uneeded columns
+        for s, t in self._transforms:
+            ops_idx, res_idx = s
+            # res_idx = _utils._as_key_indices(res, self._keys)
```
Hakuyume, line 50: Should we remove this line?

pytorch_pfn_extras/dataset/tabular/_transform.py
```diff
@@ -0,0 +1,220 @@
...
+            # We look for transformations that produces the requested keys
+            # we allow key_indices select a given key for transformations
+            # producting multiple keys since we have ensured all are disjoint
+            contained = set(res_idx).intersection(key_indices)
+            # res_idx holds the transf. indexes that belong to key_indices
+            res_idx = list(contained)
+            if len(contained) > 0:
+                # Now look the indices of the keys we need to fetch
+                # from the original dataset to apply this transformation
+                # ops_idx = _utils._as_key_indices(ops, self._dataset.keys)
```
Hakuyume, line 62: ditto.

pytorch_pfn_extras/dataset/tabular/_transform.py
emcastillo: yes! let me remove them

kmaehashi:
pfnCI, test this please.

PR #10, opened 2020-05-16
Fix example code

regonn:
The sample code executes a process using Cuda. So I fixed it to use argument information.


kmaehashi:
Thanks, LGTM!

PR #13, opened 2020-05-20
Report means in `IgniteEvaluator`

emcastillo:
When using `IgniteEvaluator` with custom metrics instead of ignite provided ones, the value reported was the last one instead of the average of the values.

The reporting scope was not correctly set.

### DIFF
--- a/pytorch_pfn_extras/training/extensions/evaluator.py
+++ b/pytorch_pfn_extras/training/extensions/evaluator.py
@@ -277,8 +277,9 @@ class IgniteEvaluator(Evaluator):
 
     def set_evaluator_handlers(self):
         from ignite.engine import Events
+
         # Register handlers to retrieve the Average metrics and report them
-        @self.evaluator.on(Events.EPOCH_STARTED)
+        @self.evaluator.on(Events.ITERATION_STARTED)
         def set_evaluation_started(engine):
             self.observation = {}
             self.cm = reporting.report_scope(self.observation)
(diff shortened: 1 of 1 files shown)


asi1024:
LGTM.

belltailjp:
Hi.
I noticed that this PR has introduced a breaking change.
PrintReport and LogReport cannot observe the metrics of evaluation.
You can reproduce it by `example/ignite-mnist.py` as follows.

```bash
# With v0.2.0 -> No problem
$ python example/ignite-mnist.py
epoch       iteration   train/loss  lr          model/fc2.bias/grad/min  val/loss    val/acc                                              
1           938         0.873671    0.01        -0.0421613               0.176294    0.9474      
2           1876        0.353082    0.01        -0.0283207               0.1115      0.9651                                           
3           2814        0.276606    0.01        -0.0249027               0.0886682   0.9725 
...

# With master (actually, after this PR is merged) -> val/loss, val/acc are not visible
$ python example/ignite-mnist.py
epoch       iteration   train/loss  lr          model/fc2.bias/grad/min  val/loss    val/acc                                                                               
1           938         0.831899    0.01        -0.0417566                                                                                                                 
2           1876        0.384308    0.01        -0.0300065                                                                                                                 
3           2814        0.309362    0.01        -0.0262319                                                                                                                 
4           3752        0.271124    0.01        -0.0249945
...
```

`result/log` that this script creates includes "val/loss" and "val/acc" with v0.2.0 ppe, while with current master it doesn't.

emcastillo:
Thanks for the report, I think is because the reporting scope, I will fix it this saturday

PR #15, opened 2020-05-20
Bugfix extended sequential

### LINKED ISSUE #14: Bug in sequential repeat when the layer has no parameters
**problems Statement**
In `init` mode, the repeated layer will be reset. In `Pytorch`, we used `reset_parameters` function to reset the parameters of layers, as here:
https://github.com/pfnet/pytorch-pfn-extras/blob/92dad97e3a3eaafb9eb2b2b804266b5ef922e4c7/pytorch_pfn_extras/nn/modules/extended_sequential.py#L15

However, there are layers have no parameters nor `reset_parameters`, such as `torch.nn.ReLU`. An error is raised when the model contains such layer.

**Error Message**
```
pytorch_pfn_extras/nn/modules/extended_sequential.py:68: in repeat
    model_list.append(self._copy_model(mode))
pytorch_pfn_extras/nn/modules/extended_sequential.py:27: in _copy_model
    return _reset_parameters(copy.deepcopy(self))
pytorch_pfn_extras/nn/modules/extended_sequential.py:9: in _reset_parameters
    _reset_parameters(submodel)
pytorch_pfn_extras/nn/modules/extended_sequential.py:17: in _reset_parameters
    model.reset_parameters()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = ReLU(), name = 'reset_parameters'

    def __getattr__(self, name):
        if '_parameters' in self.__dict__:
            _parameters = self.__dict__['_parameters']
            if name in _parameters:
                return _parameters[name]
        if '_buffers' in self.__dict__:
            _buffers = self.__dict__['_buffers']
            if name in _buffers:
                return _buffers[name]
        if '_modules' in self.__dict__:
            modules = self.__dict__['_modules']
            if name in modules:
                return modules[name]
        raise AttributeError("'{}' object has no attribute '{}'".format(
>           type(self).__name__, name))
E       AttributeError: 'ReLU' object has no attribute 'reset_parameters'
```

belldandyxtq:
This PR fixes a bug when resetting layers. In pytorch,
there are layers with no parameters and therefore no `reset_parameters`,
such as `torch.nn.ReLU`. We need to skip such layers when resetting.

This closes #14

### DIFF
--- a/pytorch_pfn_extras/nn/modules/extended_sequential.py
+++ b/pytorch_pfn_extras/nn/modules/extended_sequential.py
@@ -17,9 +17,7 @@ def _reset_parameters(model):
         elif hasattr(model, '_reset_parameters'):
             model._reset_parameters()
         else:
-            module_name = model.__module__
-            if not (module_name.startswith('torch.nn') or
-                    module_name.startswith('pytorch_pfn_extras.nn')):
+            if len(list(model.parameters())) != 0:
                 warnings.warn('Cannot reset the parameters of module {}. '
                               'Consider adding `reset_parameters` or'
                               '`_reset_parameters`'
@@ -52,7 +50,7 @@ class ExtendedSequential(torch.nn.Sequential):
         The functions is supposed to behave the same way as `repeat`
         in `chainer`.
 
-        For user-defined module, add a ``reset_parameters`` or
+        For user-defined module with parameters, add a ``reset_parameters`` or
         ``_reset_parameters`` function to repeat with mode ``init``.
         Otherwise, a warning message is generated.
 
--- a/pytorch_pfn_extras/nn/modules/extended_sequential.py
+++ b/pytorch_pfn_extras/nn/modules/extended_sequential.py
@@ -15,9 +15,6 @@ def _reset_parameters(model):
             model.reset_parameters()
         elif hasattr(model, '_reset_parameters'):
             model._reset_parameters()
-        else:
-            raise RuntimeError(
-                "Can not reset parameters of model {}".format(model))
     return model
 
 
--- a/pytorch_pfn_extras/nn/modules/extended_sequential.py
+++ b/pytorch_pfn_extras/nn/modules/extended_sequential.py
@@ -11,7 +11,8 @@ def _reset_parameters(model):
         for submodel in model.values():
             _reset_parameters(submodel)
     else:
-        if isinstance(model, torch.nn.Module):
+        if (isinstance(model, torch.nn.Module) and
+           hasattr(model, 'reset_parameters')):
             model.reset_parameters()
     return model
 
--- a/pytorch_pfn_extras/nn/modules/extended_sequential.py
+++ b/pytorch_pfn_extras/nn/modules/extended_sequential.py
@@ -11,9 +11,13 @@ def _reset_parameters(model):
         for submodel in model.values():
             _reset_parameters(submodel)
     else:
-        if (isinstance(model, torch.nn.Module) and
-           hasattr(model, 'reset_parameters')):
+        if hasattr(model, 'reset_parameters'):
             model.reset_parameters()
+        elif hasattr(model, '_reset_parameters'):
+            model._reset_parameters()
+        else:
+            raise RuntimeError(
+                "Can not reset parameters of model {}".format(model))
     return model
 
 
--- a/pytorch_pfn_extras/nn/modules/extended_sequential.py
+++ b/pytorch_pfn_extras/nn/modules/extended_sequential.py
@@ -52,6 +52,10 @@ class ExtendedSequential(torch.nn.Sequential):
         The functions is supposed to behave the same way as `repeat`
         in `chainer`.
 
+        For user-defined module, add a ``reset_parameters`` or
+        ``_reset_parameters`` function to repeat with mode ``init``.
+        Otherwise, a warning message is generated.
+
         Args:
             n_repeat (int): Number of times to repeat.
             mode (str): It should be either ``init``, ``copy``, or ``share``.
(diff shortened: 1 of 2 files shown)


kmaehashi:
Strictly speaking, the interface `reset_parameters` is defined in each module level, not i  `torch.nn.Module`. For example `MultiheadAttention` uses `_reset_parameters()`. `BatchNorm` may also require `reset_running_stats` to reset the state.

kmaehashi:
Discussed in DLFW team. Could you update the code as follows?

```
if the module name starts with torch.nn or ppe.nn (for lazy modules):
    if hasattr(model, `reset_parameters`):
        model.reset_parameters()
    elif hasattr(model, `_reset_parameters`):
        model._reset_parameters()
else:
    raise exception
```

In addition, could you add this limitation (user-defined modules cannot be used in ExtendedSeqneitial) to the docstring?

kmaehashi:
> `BatchNorm` may also require `reset_running_stats` to reset the state.

`reset_running_stats()` is called from `reset_parameters()`, so we don't have to care about this case.

belldandyxtq:
Only subclasses of `torch.nn.Module` are allowed to be added to the `Sequential` https://github.com/pytorch/pytorch/blob/d035d05080729c30636ff30fcc068de3c7e9badd/torch/nn/modules/module.py#L209

Therefore, I think we do not need to check 
> if the module name starts with torch.nn or ppe.nn (for lazy modules)

which allows user-defined layers to be added into `ExtendedSequential`.

I am not sure if we need to raise the exception if there is no `reset_parameters` nor `_reset_parameters`. We may can just ignore it. What do you think? @kmaehashi

kmaehashi:
> Only subclasses of torch.nn.Module are allowed to be added to the Sequential

User-defined custom module inherits from `torch.nn.Module`.
https://pytorch.org/tutorials/beginner/examples_nn/two_layer_net_module.html

Such classes reside outside of torch.nn or ppe.nn namespacee. When a user tries to add a user-defined custom module that does not implement `reset_parameters()` to ExtendedSequential with mode `init`, it should fail.

belldandyxtq:
I am a little confused here. I do not understand why we should raise exception for user-defined module instead of ignoring the reset or raise some warnings.
Isn't it too restrict to exclude user-defined module?

kmaehashi:
Silently ignoring the reset is surprising behavior for users. I think raising a warning instead of an exception sounds reasonable for me.

Btw, we can skip raising a warning when `len(module.parameters()) == 0`.

pytorch_pfn_extras/nn/modules/extended_sequential.py
```diff
@@ -11,8 +12,18 @@ def _reset_parameters(model):
         for submodel in model.values():
             _reset_parameters(submodel)
     else:
-        if isinstance(model, torch.nn.Module):
+        if hasattr(model, 'reset_parameters'):
             model.reset_parameters()
+        elif hasattr(model, '_reset_parameters'):
+            model._reset_parameters()
+        else:
+            if len(list(model.parameters())) != 0:
```
kmaehashi, line 20: Maybe it's better to check buffers as well:
`len(list(model.parameters(recurse=False))) != 0 or len(list(model.buffers(recurse=False))) != 0`

pytorch_pfn_extras/nn/modules/extended_sequential.py
```diff
@@ -39,6 +50,10 @@ def repeat(self, n_repeat: int, mode: 'str' = 'init'):
         The functions is supposed to behave the same way as `repeat`
         in `chainer`.
 
+        For user-defined module with parameters, add a ``reset_parameters`` or
```
kmaehashi, line 53: Could you elaborate more? e.g.,

> When the mode is set to `init` (this is the default), modules will be copied and reinitialized by calling `reset_parameters` (or `_reset_parameters`) method. To register user-defined modules with parameters or buffers to this Sequential, you need to implement `reset_parameters` method to the module to reinitialize parameters and (if necessary) buffers; otherwise the initialization cannot be performed and a warning message will be shown.

pytorch_pfn_extras/nn/modules/extended_sequential.py
belldandyxtq: Thank you, fixed in 34eba76

pytorch_pfn_extras/nn/modules/extended_sequential.py
belldandyxtq: Fixed in 00f32ab

kmaehashi:
Thank you! LGTM 👍

pytorch_pfn_extras/nn/modules/extended_sequential.py
```diff
@@ -11,8 +12,19 @@ def _reset_parameters(model):
...
+        if hasattr(model, 'reset_parameters'):
             model.reset_parameters()
+        elif hasattr(model, '_reset_parameters'):
+            model._reset_parameters()
+        else:
+            if (len(list(model.parameters())) != 0 or
+                    len(list(model.buffers())) != 0):
+                warnings.warn('Cannot reset the parameters of module {}. '
+                              'Consider adding `reset_parameters` or'
+                              '`_reset_parameters`'
```
kmaehashi, line 24: nitpick
```suggestion
                              'Consider adding `reset_parameters` or '
                              '`_reset_parameters` '
```

kmaehashi:
> We will raise this discussion to the PyTorch team (to make `reset_parameters` defined at `torch.nn.Module` level) to remove this restriction in the future.

For reference: https://github.com/pytorch/pytorch/pull/39051

pytorch_pfn_extras/nn/modules/extended_sequential.py
belldandyxtq: Done in 9048846

kmaehashi:
LGTM!

PR #16, opened 2020-05-21
Allow transformation functions for snapshots

emcastillo:
Follows #7

### DIFF
--- a/pytorch_pfn_extras/training/extensions/_snapshot.py
+++ b/pytorch_pfn_extras/training/extensions/_snapshot.py
@@ -142,6 +142,12 @@ n_retains=-1, autoload=False)
             Automatic loading only works when the filename is a string.
         saver_rank (int): If defined, the snapshot will be taken by only one
             rank when running in distributed mode and restored by all.
+        transform_models (callable): If defined, function to apply to a model
+            before obtaining its `state_dict`. Takes two parameters, the object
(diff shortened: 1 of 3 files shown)


pytorch_pfn_extras/training/extensions/_snapshot.py
```diff
@@ -142,6 +142,12 @@ def snapshot_object(target, filename, savefun=None, **kwargs):
             Automatic loading only works when the filename is a string.
         saver_rank (int): If defined, the snapshot will be taken by only one
             rank when running in distributed mode and restored by all.
+        transform_models (func): If defined, function to apply to a model
```
asi1024, line 145: We write `callable` conventionally in such a situation.
```suggestion
        transform_models (callable): If defined, function to apply to a model
```

emcastillo:
PTAL!

pytorch_pfn_extras/training/extensions/_snapshot.py
asi1024: Could you fix this part too?
https://github.com/pfnet/pytorch-pfn-extras/pull/16/files#diff-21bebc7a5bc35433c970655418a04770R227-R232

emcastillo:
Should be fixed now

asi1024:
LGTM.

PR #17, opened 2020-05-22
Fix ignite agregate metrics not being reported

emcastillo:
Fix the error introduced in #13 

The ignite metrics were not reported in the correct scope

### DIFF
--- a/tests/pytorch_pfn_extras_tests/training_tests/extensions_tests/test_evaluator.py
+++ b/tests/pytorch_pfn_extras_tests/training_tests/extensions_tests/test_evaluator.py
@@ -216,20 +216,6 @@ def create_dummy_evaluator(model):
     return evaluator
 
 
-class IgniteDataset(torch.utils.data.Dataset):
-    def __init__(self, n_data):
-        self.values = [i for i in range(n_data)]
-        self.n_data = n_data
-
-    def __len__(self):
-        return len(self.values)
-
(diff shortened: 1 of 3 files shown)


emcastillo:
I just added a test to verify that both, internally reported values are being correctly averaged, and that `ignite.metrics` are accessible at the reporter observation.

Thanks to @corochann for giving us the initial draft to test the ignite evaluators.

tests/pytorch_pfn_extras_tests/training_tests/extensions_tests/test_evaluator.py
```diff
@@ -188,3 +188,70 @@ def test_evaluator_progress_bar():
...
+        model()
+        # We return fake results for the reporters to
+        # and metrics to work
+        return (y_pred, y_pred)
+
+    evaluator = Engine(update_fn)
+    return evaluator
+
+
+class IgniteDataset(torch.utils.data.Dataset):
```
corochann, line 219: I noticed shorter way to write dataset is:

```
n_data = 10
x = torch.randn((n_data , 2), requires_grad=True)
y = torch.randn((n_data , 2))
dataset = TensorDataset(y, t)
```

emcastillo: Solved! thanks :D

asi1024:
LGTM.

PR #20, opened 2020-05-25
Remove `FoolUpdater` and do everything through the manager

emcastillo:
The `updater` attribute of the manager was introduced to simplify adapting some components to PyTorch. However,  since there is no real Updater in ppe, it shouldn't be relied upon.

This change might break some user code when they use their old chainer extensions with ppe but I think is a good moment to get rid of it.

### DIFF
--- a/pytorch_pfn_extras/training/extensions/_snapshot.py
+++ b/pytorch_pfn_extras/training/extensions/_snapshot.py
@@ -114,7 +114,7 @@ n_retains=-1, autoload=False)
         filename (str): Name of the file into which the object is serialized.
             It can be a format string, where the manager object is passed to
             the :meth:`str.format` method. For example,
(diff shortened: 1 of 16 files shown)


emcastillo:
I would like to ask the opinion of @corochann and @belltailjp about removing the updater 😇

belltailjp:
I think it should be fine, as the one I proposed to introduce itself (#18) doesn't (and perhaps won't) manage these info. For my codes currently under dev/maintenance, migration cost is affordable.

corochann:
I'm fine with the change too.

kmaehashi:
pfnCI, test this please.

kmaehashi:
Could you check the test failure?

emcastillo:
rebased!

kmaehashi:
pfnCI, test this please.

kmaehashi:
LGTM!

PR #22, opened 2020-05-27
Not to print "not found" in snapshot initialize, return the loaded path instead in autoload mode

belltailjp:
`_Snapshot.initialize` with `autoload=True` prints the following message to stdout when there's no file to load.

```
No snapshot file that matches last_snapshot was found
```

In multi-GPU setting using MPI and torch.distributed, it leads to occupying the terminal screen with its repetition.

```
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
No snapshot file that matches last_snapshot was found
...
```

I know it is useful to let users know if the snapshot has autoloaded a snapshot, so I introduce a way to more programmatically notify it, so users can handle it as they want.
In summary this PR includes:

* Suppress stdout print.
* `snapshot.initialize` to return the file path that is actually loaded, or None if nothing (including the case the snapshot is `autoload=False`)

Although this should technically work, my concern is that auto-loading process might not be only the thing that `_Snapshot.initialize` is in charge in the future (although currently it almost is), so reserving its return value by the loaded filename could cause of future headache.

### DIFF
--- a/tests/pytorch_pfn_extras_tests/training_tests/extensions_tests/test_snapshot.py
+++ b/tests/pytorch_pfn_extras_tests/training_tests/extensions_tests/test_snapshot.py
@@ -132,9 +132,21 @@ def test_multi_target_autoload(remover):
     snapshot2 = extensions.snapshot_object(target, 'myfile.dat',
                                            autoload=True)
     # Load the snapshot and verify it
-    snapshot2.initialize(new_trainer)
+    assert snapshot2.initialize(new_trainer) == 'myfile.dat'
     assert new_trainer.state_dict() == trainer.state_dict()
     assert new_other.state_dict() == other_state_dict
+
+
+def test_multi_target_autoload_not_found(remover):
+    trainer = get_trainer_with_mock_updater(out_dir='.')
+    other = _StateDictObj(state_dict={'original': 'state'})
+
+    target = {'trainer': trainer, 'other': other}
+    snapshot = extensions.snapshot_object(target, 'myfile.dat',
+                                          autoload=True)
+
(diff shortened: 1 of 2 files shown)


tests/pytorch_pfn_extras_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -132,11 +132,23 @@ def test_multi_target_autoload(remover):
...
                                            autoload=True)
     # Load the snapshot and verify it
-    snapshot2.initialize(new_trainer)
+    assert snapshot2.initialize(new_trainer) == 'myfile.dat'
     assert new_trainer.state_dict() == trainer.state_dict()
     assert new_other.state_dict() == other_state_dict
 
 
+def test_multi_target_autoload_not_found(remover):
+    trainer = get_trainer_with_mock_updater(out_dir='.')
```
emcastillo, line 141: The CI complains that this is not valid, but is actually valid and  runs well in local env!?

emcastillo: rebased & pushed

PR #23, opened 2020-05-28
Avoid multiple writes and prints on distributed settings

emcastillo:
`ExtensionsManager` now holds the distributed `rank` information.

Some extensions that write to files or print to stdout do that action only in the master rank. However, they can be executed in the worker ranks without side-effects.

There are some issues such as, if in a distributed environment, maybe there is some scenario where we want the workers to write to a specific folder each, and abstract that through a writer. So I wonder if enforcing an only `rank==0` writing in the code base is good.

### DIFF
--- a/pytorch_pfn_extras/training/manager.py
+++ b/pytorch_pfn_extras/training/manager.py
@@ -122,8 +122,12 @@ class _BaseExtensionsManager:
         for ext in extensions:
             self.extend(ext)
 
-        # Initialize the writer
-        self.writer.initialize(self.out)
+        # Get track of the process ranks
+        self.size, self.rank, self.local_rank = _get_ranks_from_env()
+
+        # Initialize the writer on the master rank
+        if self.rank == 0:
+            self.writer.initialize(self.out)
 
     @property
     def elapsed_time(self):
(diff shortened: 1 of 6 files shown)


pytorch_pfn_extras/training/manager.py
```diff
@@ -122,8 +122,12 @@ def __init__(
...
             self.extend(ext)
 
-        # Initialize the writer
-        self.writer.initialize(self.out)
+        # Get track of the process ranks
+        self.size, self.rank, self.local_rank = _get_ranks_from_env()
+
+        # Initialize the writer on the master rank
+        if self.rank == 0:
+            self.writer.initialize(self.out)
```
emcastillo, line 130: This might not be correct at all. Needs discussion

kmaehashi:
I think it is a responsibility of users to decide which extensions need to be activated on each node, like as designed in Chainer:

https://github.com/chainer/chainer/blob/8b9e9c014716077366845fdb5bf780644c067952/examples/chainermn/mnist/train_mnist.py#L114-L116

We need to document and/or add examples this point.

kmaehashi:
@emcastillo Do you think we can close this PR?

PR #32, opened 2020-06-03
Fix dist snapshot

### LINKED ISSUE #29: `BestValueTrigger` does not work well in mpi run like multi node setting
## What happened
`_DistributedSnapshot` with `BestValueTrigger` gets stuck.

### code
https://gist.github.com/dhgrs/56424106e00bafee9617b0a15a028c2c

### command
`CUDA_VISIBLE_DEVICES=0,1 mpiexec -N 2 python3 mnist.py`

## Why it causes
Reporter works in all mpi processes without all reduce operation so that `BestValueTrigger` check different values in each process. It causes that some processes are triggered but the others are not. `_DistributedSnapshot` waits for all mpi processes but some of them would never finish because are not triggered.

## Workaround
Apply all reduce operation manually before reporting. But is this the best way? Will ppe support auto all reduce?

emcastillo:
Fixes #29 

Synchronization is implicitly done when the weights are reduced and aggregated. So the barrier can be safely removed.

Also, the current extension only allows one worker to save the snapshot when the trigger is fired.
Now when the `saver_rank` is set to -1, all the ranks will save their status.

TODO: improve test

### DIFF
--- a/pytorch_pfn_extras/training/extensions/_snapshot.py
+++ b/pytorch_pfn_extras/training/extensions/_snapshot.py
@@ -215,6 +215,7 @@ def snapshot(savefun=None,
             by :func:`torch.save` .
         saver_rank (int): If defined, the snapshot will be taken by only one
             rank when running in distributed mode and restored by all.
+            If -1 is specified, all the workers will save the snapshot.
     Returns:
         Snapshot extension object.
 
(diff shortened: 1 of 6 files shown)


kmaehashi:
Converted to the draft PR until TODOs resolved.

PR #33, opened 2020-06-03
Add `TensorBoardWriter`

### LINKED ISSUE #28: Do you have a plan to support TensorBoard
PyTorch supports TensorBoard naitively in `torch.utils.tensorboard`. But ppe's `LogReport` doesn't because tensorbord writers don't satisfy the requirement.

https://github.com/pfnet/pytorch-pfn-extras/blob/b2ddc4570f655bfe00f1d9f7dbb64fd4654a73c4/pytorch_pfn_extras/training/extensions/log_report.py#L57-L60

So my question is would ppe support it? Or do we have any other way to do it?

emcastillo:
Closes #28 

Creating the `LogReport` extension with the kwarg `writer=TensorBoardWriter()`
should automatically report the metrics to tensorboard.

Users can also use this writer to report their own metrics in their modules.


emcastillo:
Thanks a lot to @belltailjp for his help when doing the writing system, this allowed us to solve this pretty easily.

pytorch_pfn_extras/writing.py
```diff
@@ -500,3 +500,21 @@ def create_queue(self):
...
     def create_consumer(self, q):
         return multiprocessing.Process(target=self.consume, args=(q,))
+
+
+class TensorBoardWriter(object):
+    def __init__(self, savefun=None, fs=None, out_dir=None, **kwds):
+        import torch.utils.tensorboard
+        self._writer = torch.utils.tensorboard.SummaryWriter()
+
+    def __call__(self, filename, out_dir, target, *, savefun=None):
```
emcastillo, line 510: __call__ and __init__ attributes are just to keep compatibility.

emcastillo:
I tested it with tensorboard and I confirm it works.
Changed the `ppe.training.extensions.LogReport()` in the mnist example to
`ppe.training.extensions.LogReport(writer=ppe.writing.TensorBoardWriter())`

pytorch_pfn_extras/writing.py
```diff
@@ -500,3 +500,21 @@ def create_queue(self):
 
     def create_consumer(self, q):
         return multiprocessing.Process(target=self.consume, args=(q,))
+
+
+class TensorBoardWriter(object):
+    def __init__(self, savefun=None, fs=None, out_dir=None, **kwds):
+        import torch.utils.tensorboard
+        self._writer = torch.utils.tensorboard.SummaryWriter()
```
dhgrs, line 508: `torch.utils.tensorboard.SummaryWriter()` is initialized with the default arguments. So the log file is saved in `runs/CURRENT_DATETIME_HOSTNAME` as described in the document.
https://pytorch.org/docs/stable/tensorboard.html#torch.utils.tensorboard.writer.SummaryWriter

How about use `out_dir` to initialize it?

emcastillo:
Just pushed a change with support for `out_dir` and directly specifying the keyword args to the underlying `SummaryWriter`, also a test was added

kmaehashi:
I confirmed that `tensorboard --logdir runs` correctly displayed the graph.

PR #35, opened 2020-06-17
Add `step_optimizers` parameter to `Manager.run_iteration`

emcastillo:
The manager can zero the gradients and call the step at the end for the selected optimizers


emcastillo:
PTAL!

asi1024:
LGTM!

PR #42, opened 2020-07-01
Fix example for new ignite version

emcastillo:
ignite 0.4.0 errored because the model was not being transferred to the gpu.


kmaehashi:
LGTM!

PR #43, opened 2020-07-06
Add a forked `DataLoader` with reusable workers

emcastillo:
The DataLoader code is forked from PyTorch and it contains my changes in https://github.com/pytorch/pytorch/pull/35795/files

The idea is to get a DataLoader interface here to implement our changes and distribute them while the changes are being upstreamed to PyTorch.


kmaehashi:
/test

@emcastillo Could you add docs? (I think simple one is enough)

kmaehashi:
@emcastillo Could you check style check failures?

PR #50, opened 2020-07-16
Added LazyBatchNorm(1,2,3)d

### LINKED ISSUE #48: Feature request: LazyBatchNorm
In my project, I implemented LazyBatchNorm like this and I think pytorch-pfn-extras should have the similar feature:

```
class LazyBatchNorm1d(
    ppe.nn.modules.lazy.LazyInitializationMixin, torch.nn.BatchNorm1d
):
    """
    LazyBatchNorm1d is a lazy version of BatchNorm1d.  It does not require to
    specify the number of the input parameters beforehand.
    """

    lazy_parameter_names = ("weight", "bias")

    def __init__(self, num_features, *args, **kwargs):
        super().__init__(num_features or 0, *args, **kwargs)
        if num_features is None:
            self.num_features = None
            self.weight = ppe.nn.modules.lazy.UninitializedParameter()
            self.bias = ppe.nn.modules.lazy.UninitializedParameter()

    def forward(self, input):
        if isinstance(self.weight, ppe.nn.modules.lazy.UninitializedParameter):
            self.num_features = input.shape[-1]
            if self.affine:
                self.weight = torch.nn.Parameter(
                    self.weight.new_empty(self.num_features)
                )
                self.bias = torch.nn.Parameter(self.weight.new_empty(self.num_features))
            if self.track_running_stats:
                self.running_mean = torch.zeros(
                    self.num_features, device=self.running_mean.device
                )
                self.running_var = torch.ones(
                    self.num_features, device=self.running_var.device
                )
            self.reset_parameters()
        return super().forward(input)

    def reset_parameters(self):
        if self.lazy_parmeters_determined:
            super().reset_parameters()
```

Another discussion:
It would be good to have LazyBatchNorm instead of having LazyBatchNorm*d-s since lazy ones know its desired shapes. Any thoughts?

emcastillo:
Closes #48 

Based on the implementation by @imos

### DIFF
--- a/pytorch_pfn_extras/nn/modules/lazy_batchnorm.py
+++ b/pytorch_pfn_extras/nn/modules/lazy_batchnorm.py
@@ -4,8 +4,9 @@ from pytorch_pfn_extras.nn.modules.lazy import LazyInitializationMixin
 from pytorch_pfn_extras.nn.modules.lazy import UninitializedParameter
 
 
-class _LazyNormBase(
-    LazyInitializationMixin, torch.nn.modules.batchnorm._NormBase
+class _LazyBatchNorm(
+    LazyInitializationMixin,
+    torch.nn.modules.batchnorm._BatchNorm
 ):
 
     lazy_parameter_names = ('weight', 'bias')
@@ -48,11 +49,6 @@ class _LazyNormBase(
                     dtype=self.running_mean.dtype
                 )
             self.reset_parameters()
-        return super().forward(input)
-
-
-class _LazyBatchNorm(_LazyNormBase, torch.nn.modules.batchnorm._BatchNorm):
-    def forward(self, input):
         return super().forward(input)
 
 
--- a/pytorch_pfn_extras/nn/modules/lazy_batchnorm.py
+++ b/pytorch_pfn_extras/nn/modules/lazy_batchnorm.py
@@ -57,12 +57,27 @@ class _LazyBatchNorm(_LazyNormBase, torch.nn.modules.batchnorm._BatchNorm):
 
 
 class LazyBatchNorm1d(_LazyBatchNorm, torch.nn.BatchNorm1d):
+    """BatchNorm1d module with lazy weight initialization.
+
+    When ``num_features`` is ``None``, it is determined at the first time of
+    the forward step.
+    """
     pass
 
 
 class LazyBatchNorm2d(_LazyBatchNorm, torch.nn.BatchNorm2d):
+    """BatchNorm2d module with lazy weight initialization.
+
+    When ``num_features`` is ``None``, it is determined at the first time of
+    the forward step.
+    """
     pass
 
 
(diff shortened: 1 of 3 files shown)


pytorch_pfn_extras/nn/modules/lazy_batchnorm.py
```diff
@@ -0,0 +1,83 @@
...
+                self.running_var = torch.ones(
+                    self.num_features, device=self.running_var.device,
+                    dtype=self.running_mean.dtype
+                )
+            self.reset_parameters()
+        return super().forward(input)
+
+
+class _LazyBatchNorm(_LazyNormBase, torch.nn.modules.batchnorm._BatchNorm):
+    def forward(self, input):
```
kmaehashi, line 55: Is this override needed?

kmaehashi (changes requested):

pytorch_pfn_extras/nn/modules/lazy_batchnorm.py
```diff
@@ -0,0 +1,83 @@
+import torch
+
+from pytorch_pfn_extras.nn.modules.lazy import LazyInitializationMixin
+from pytorch_pfn_extras.nn.modules.lazy import UninitializedParameter
+
+
+class _LazyNormBase(
+    LazyInitializationMixin, torch.nn.modules.batchnorm._NormBase
+):
```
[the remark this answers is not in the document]
emcastillo: Both of them implement forward so we need to call both 😅

pytorch_pfn_extras/nn/modules/lazy_batchnorm.py
emcastillo: This is how it was implemented in torch, there is a general forward to all the batchnorm implementations, and then dimension specific ones.

pytorch_pfn_extras/nn/modules/lazy_batchnorm.py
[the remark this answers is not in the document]
emcastillo: torch.nn.modules.batchnorm._NormBase holds all the batchnorm generic code, I am not overriding everything so I keep it as the base class.
I think ppe lazyconv modules are implemented like this.

emcastillo:
PTAL

pytorch_pfn_extras/nn/modules/lazy_batchnorm.py
```diff
@@ -0,0 +1,79 @@
...
+                self.running_mean = torch.zeros(
+                    self.num_features, device=self.running_mean.device,
+                    dtype=self.running_mean.dtype
+                )
+                self.running_var = torch.ones(
+                    self.num_features, device=self.running_var.device,
+                    dtype=self.running_mean.dtype
+                )
+            self.reset_parameters()
+        return super().forward(input)
```
emcastillo, line 52: This calls `torch.nn.modules.batchnorm._BatchNorm.forward` so I overcomplicated it because I got a bit confused.

kmaehashi:
LGTM!

PR #55, opened 2020-08-24
Fix resuming `Summary` reporting

colspan:
This PR suggests to fix the resume procedure of `Summary` class.

I got a `RuntimeError` when resuming from some snapshots.
```
  File "./lib/python3.8/site-packages/pytorch_pfn_extras/training/extensions/log_report.py", line 90, in __call__
    summary.add(observation)
  File "./lib/python3.8/site-packages/pytorch_pfn_extras/reporting.py", line 342, in add
    summaries[k].add(v, weight=w)
  File "./lib/python3.8/site-packages/pytorch_pfn_extras/reporting.py", line 270, in add
    self._x += weight * value
RuntimeError: a leaf Variable that requires grad has been used in an in-place operation.
```

So I simply patch `Summary.load_state_dict` method as follows.
```Python
def load_state_dict(self, to_load):
        self._x = torch.tensor(to_load['_x'], requires_grad = False)
        self._x2 = torch.tensor(to_load['_x2'], requires_grad = False)
        self._n = torch.tensor(to_load['_n'], requires_grad = False)
```

But I still got warnings like 
```
UserWarning: To copy construct from a tensor,
        it is recommended to use sourceTensor.clone().detach() or
        sourceTensor.clone().detach().requires_grad_(True),
        rather than torch.tensor(sourceTensor).
```

Currently I'm not familiar with PyTorch, so it's hard to propose an appropriate solution for this issue.
I would be really grateful if anyone could discuss and solve this.

Thank you.

### DIFF
--- a/pytorch_pfn_extras/reporting.py
+++ b/pytorch_pfn_extras/reporting.py
@@ -303,13 +303,9 @@ class Summary:
         return state
 
     def load_state_dict(self, to_load):
-        # TODO : still got warning `UserWarning: To copy construct from a tensor,
-        #        it is recommended to use sourceTensor.clone().detach() or
-        #        sourceTensor.clone().detach().requires_grad_(True),
-        #        rather than torch.tensor(sourceTensor).
-        self._x = torch.tensor(to_load['_x'], requires_grad = False)
-        self._x2 = torch.tensor(to_load['_x2'], requires_grad = False)
-        self._n = torch.tensor(to_load['_n'], requires_grad = False)
+        self._x = numpy.float64(to_load['_x'])
+        self._x2 = numpy.float64(to_load['_x2'])
+        self._n = numpy.float64(to_load['_n'])
 
 class DictSummary:
 
--- a/pytorch_pfn_extras/reporting.py
+++ b/pytorch_pfn_extras/reporting.py
@@ -303,10 +303,13 @@ class Summary:
         return state
 
     def load_state_dict(self, to_load):
-        self._x = to_load['_x']
-        self._x2 = to_load['_x2']
-        self._n = to_load['_n']
-
+        # TODO : still got warning `UserWarning: To copy construct from a tensor,
+        #        it is recommended to use sourceTensor.clone().detach() or
+        #        sourceTensor.clone().detach().requires_grad_(True),
+        #        rather than torch.tensor(sourceTensor).
+        self._x = torch.tensor(to_load['_x'], requires_grad = False)
+        self._x2 = torch.tensor(to_load['_x2'], requires_grad = False)
+        self._n = torch.tensor(to_load['_n'], requires_grad = False)
 
 class DictSummary:
 
--- a/tests/pytorch_pfn_extras_tests/test_reporter.py
+++ b/tests/pytorch_pfn_extras_tests/test_reporter.py
@@ -223,6 +223,19 @@ def test_summary_weight():
     numpy.testing.assert_allclose(mean.numpy(), val)
 
 
+def test_summary_resume():
+    summary = ppe.reporting.Summary()
+    summary.add(torch.tensor([[15.0]], requires_grad=True))
+
+    f = io.BytesIO()
+    torch.save(summary.state_dict(), f)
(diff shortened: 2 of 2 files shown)


kmaehashi:
Minimum reproducer:

```py
import torch

s = ppe.reporting.Summary()
s.add(torch.tensor([[10.0]], requires_grad=True))
torch.save(s.state_dict(), 'out.pth')
s.load_state_dict(torch.load('out.pth'))
s.add(10)
```

I guess it's better to convert loaded values as scalar? @emcastillo

kmaehashi:
We discussed and it seems it's better to convert the value as numpy scalar instead of Python scalar nor torch.Tensor. Could you work on it or shall we takeover this?

colspan:
Thank you for reviewing and discussion. I tried to apply your suggestion.
Please have a quick review on this.

pytorch_pfn_extras/reporting.py
```diff
@@ -303,10 +303,9 @@ def state_dict(self):
         return state
 
     def load_state_dict(self, to_load):
-        self._x = to_load['_x']
-        self._x2 = to_load['_x2']
-        self._n = to_load['_n']
-
+        self._x = numpy.float64(to_load['_x'])
```
kmaehashi, line 306: I think `numpy.asarray` is better to preserve dtype.
Note that values inside `to_load` maybe any of Python float (`1.0`), NumPy array or torch.Tensor...

colspan: I tried to replace `numpy.float64` with `numpy.asarray`, but I got `TypeError: can't convert CUDA tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.`.


So, I prepared a type aware converter method like follows.

```Python
def load_as_numpy_array(x):
    # return numpy.asarray(x)
    if isinstance(x, torch.Tensor):
        return torch.detach(x).cpu()#.numpy()
    else:
        return numpy.asarray(x)
```


```Python
    def load_state_dict(self, to_load):
        self._x = load_as_numpy_array(to_load['_x'])
        self._x2 = load_as_numpy_array(to_load['_x2'])
        self._n = load_as_numpy_array(to_load['_n'])
```

But still `TypeError` occurs for some metrics.

```
  File "./lib/python3.8/site-packages/pytorch_pfn_extras/training/extensions/log_report.py", line 90, in __call__
    summary.add(observation)
  File "./lib/python3.8/site-packages/pytorch_pfn_extras/reporting.py", line 350, in add
    summaries[k].add(v, weight=w)
  File "./lib/python3.8/site-packages/pytorch_pfn_extras/reporting.py", line 272, in add
    self._x += weight * value
TypeError: unsupported operand type(s) for +=: 'numpy.ndarray' and 'Tensor'
```

Here is a reproducer.

```Python
    summary = ppe.reporting.Summary()
    summary.add(torch.tensor([[15.0]], requires_grad=True))

    f = io.BytesIO()
    torch.save(summary.state_dict(), f)
    summary.load_state_dict(torch.load(io.BytesIO(f.getvalue())))

    summary.add(torch.tensor(5.0)) # torch.tensor input
    mean = summary.compute_mean() # => TypeError: unsupported operand type(s) for +=: 'numpy.ndarray' and 'Tensor'
    numpy.testing.assert_allclose(mean, 10)
```

Finally this patch solves this issue.

```Python
    def add(self, value, weight=1):
        """Adds a scalar value.

        Args:
            value: Scalar value to accumulate. It is either a NumPy scalar or
                a zero-dimensional array (on CPU or GPU).
            weight: An optional weight for the value. It is a NumPy scalar or
                a zero-dimensional array (on CPU or GPU).
                Default is 1 (integer).

        """
        value = load_as_numpy_array(value) #### <- patched
        self._x += weight * value
        self._x2 += weight * value * value
        self._n += weight
```

I feel these patches are too naive, so I would like to find additional ideas to make this simple.

Thank you.

kmaehashi: Sorry for the delayed reply  :bow: I will fix and and add tests in #60.

kmaehashi:
Fixed in #60.

PR #57, opened 2020-08-24
Treat warnings as error

kmaehashi:


kmaehashi:
Fixed to avoid copying tensor using `torch.tensor`. `SystemError` issue (PyTorch bug) will be fixed in https://github.com/pytorch/pytorch/pull/44371

emcastillo (approved):

take-cheeze:
As commented in https://github.com/pfnet/pytorch-pfn-extras/pull/64/files#issuecomment-689993407 following test cases still fail:
```
FAILED tests/pytorch_pfn_extras_tests/onnx/test_annotate.py::test_annotate - DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will sto...
FAILED tests/pytorch_pfn_extras_tests/onnx/test_annotate.py::test_apply_annotation - DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it ...
FAILED tests/pytorch_pfn_extras_tests/onnx/test_annotate.py::test_scoped_anchor - SystemError: <method '__bool__' of 'torch._C._TensorBase' objects> returned a result with an error set
FAILED tests/pytorch_pfn_extras_tests/onnx/test_export_testcase.py::test_backward_multiple_input - SystemError: <method '__bool__' of 'torch._C._TensorBase' objects> returned a result with an error set
```
It's caused by ONNX and PyTorch so maybe these need to be ignored

take-cheeze:
Seems this was closed with comment of https://github.com/pfnet/pytorch-pfn-extras/pull/64

PR #61, opened 2020-09-08
Add ProgressBarNotebook

corochann:
`ProgressBarNotebook` class is useful when you work with jupyter notebook.
The functionality is originally implemented in [grafi-tt/chaineripy](https://github.com/grafi-tt/chaineripy) by @grafi-tt as chainer Traner's extension.
I modified it to work on current ppe.

Example usage can be found [here](https://www.kaggle.com/corochann/lyft-training-with-multi-mode-confidence).


pytorch_pfn_extras/training/extensions/__init__.py
```diff
@@ -7,6 +7,7 @@
 from pytorch_pfn_extras.training.extensions.micro_average import MicroAverage  # NOQA
 from pytorch_pfn_extras.training.extensions.print_report import PrintReport  # NOQA
 from pytorch_pfn_extras.training.extensions.progress_bar import ProgressBar  # NOQA
+from pytorch_pfn_extras.training.extensions.progress_bar_notebook import ProgressBarNotebook  # NOQA
```
corochann, line 10: This might add additional software dependency to IPython or ipywidgets.
Is it better to avoid that and we should import it as
`from pytorch_pfn_extras.training.extensions.progress_bar_notebook import ProgressBarNotebook as ProgressBar` ??

pytorch_pfn_extras/training/extensions/progress_bar_notebook.py
```diff
@@ -0,0 +1,140 @@
...
+            :class:`IntervalTrigger`, this extension uses its attributes to
+            determine the length of the training.
+        update_interval (int): Number of iterations to skip printing the
+            progress bar.
+        bar_length (int): This is not used, argument is kept to be consistent with `ProgressBar`.
+        out: This is not used, argument is kept to be consistent with `ProgressBar`.
+    """
+
+    def __init__(self, training_length=None, update_interval=100,
+                 bar_length=50, out=sys.stdout):
```
corochann, line 32: I kept unused 2 arguments to switch between ProgressBar & ProgressBarNotebook easily. what do you think?

pytorch_pfn_extras/training/extensions/__init__.py
emcastillo: Lets wrap it in a try/except

corochann: Wrapped by try except

corochann:
Which way do you want to adopt?

1. Install ipython library in setup.py, it will force user to install ipython library during ppe installation
2. Install ipython library in test: https://github.com/pfnet/pytorch-pfn-extras/blob/master/.flexci/test.sh#L7
3. Skip pytest when ipython library is not installed and module is not imported

corochann:
May I adopt the option 2? @emcastillo

emcastillo:
Let's do both 2 and 3, install it in test.sh and skip it if is not present (for local development)
Thanks!

corochann:
@emcastillo test passed, please review.

emcastillo (approved):

PR #71, opened 2020-09-24
Fix extensions using matplotlib to emit warning only once

kmaehashi:
This fixes warnings to emit warnings only once in these extensions so that users won't see too many warnings when using `python -Wall`.

### DIFF
--- a/pytorch_pfn_extras/training/extensions/variable_statistics_plot.py
+++ b/pytorch_pfn_extras/training/extensions/variable_statistics_plot.py
@@ -260,7 +260,7 @@ grid=True)
         return _available
 
     def __call__(self, manager):
-        if self.available():
+        if _available:
             # Dynamically import pyplot to call matplotlib.use()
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/training/extensions/variable_statistics_plot.py
```diff
@@ -260,7 +260,7 @@ def available():
         return _available
 
     def __call__(self, manager):
-        if self.available():
+        if _available:
```
emcastillo, line 263: Availability check is not performed? `_check_available()`

kmaehashi: I overlooked that `VariableStaticsPlot` didn't call it in `__init__`. Fixed!

emcastillo: Thanks!!! 🙇‍♂️

emcastillo (approved):

PR #74, opened 2020-09-25
Rewrite config documentation

yuyu2172:
Current documentation does not cover entire funtionality of Config system.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1,11 +1,11 @@
 - [Config](#config)
   - [Basic](#basic)
-  - [Interpolation](#interpolation)
-    - [Callable Interpolation](#callable-interpolation)
-    - [Interpolation by Path](#interpolation-by-path)
-    - [Interpolation by Attribute](#interpolation-by-attribute)
-    - [Default Value by Path Interpolation](#default-value-by-path-interpolation)
-    - [Ignore Interpolation](#ignore-interpolation)
+  - [Substitution](#substitution)
+    - [Callable Substitution](#callable-substitution)
+    - [Substitution by Path](#substitution-by-path)
+    - [Substitution by Attribute](#substitution-by-attribute)
+    - [Default Value by Path Substitution](#default-value-by-path-substitution)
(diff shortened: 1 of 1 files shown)


kmaehashi:
@Hakuyume Could you review this PR? Thanks :pray:

docs/config.md
```diff
@@ -1,100 +1,272 @@
-# Config system
+- [Config](#config)
+  - [Basic](#basic)
+  - [Interpolation](#interpolation)
```
Hakuyume, line 3: How about using "substitution" instead of "interpolation"?

docs/config.md
```diff
@@ -1,100 +1,272 @@
...
-Algorithm configurations such as training hyperparameters are serialized as nested mappings often using the YAML format.  
-Nested mappings are parsed by `pytorch_pfn_extras.config.Config` that returns a dictionary of objects that are lazily evaluated.
-The most notable feature is a special handling of the `type` key, which invokes a function or instantiates an object. The arguments for them are mapped by non-`type` keys.
-```python 
-import yaml
+
+# Config
+
+## Basic
+Config object is created from a dictionary.
```
Hakuyume, line 15: Config also supports a list, a string, and a number.

docs/config.md
```diff
@@ -1,100 +1,272 @@
...
+
 config = Config(pre_eval_config, types)
-resnet = config['/model']  # same as resnet50(num_classes=10)
+# the value returned by
+# concat(x0='First', x1='Last')
+print(config['/name'])
+# 'First Last'
+```
+
+#### How it works
```
Hakuyume, line 71: Do we need this section? `resolve_key` is inaccurate and not helpful to understand the behavior. I think `the value returned by concat(x0='First', x1='Last')` (in the example) is enough to explain.

Hakuyume:
@yuyu2172 @kmaehashi I commented.

docs/config.md
yuyu2172: I deleted this line

PR #77, opened 2020-10-07
Support built-in config types for Optuna

dhgrs:
ppe's config supports custom types. It seems nice to prepare built-in types for related libraries like Optuna. In this PR, I added 2 features, `optuna_types` and `load_path_with_optuna_types`.

## `optuna_types`
It returns types dictionary according to input trial. We need to update it when Optuna adds new types.

## `load_path_with_optuna_types`
It's similar to `Config.load_path`. The difference is default types. It adds `suggest_*` according to input trial. If input types already has `suggest_*` key, it warns.

---

In `pytorch_pfn_extras/config_types.py`, Optuna is never imported so that it does not require to install Optuna if user doesn't use it. But for test, it requires Optuna. Then `setup.py` is updated.

### DIFF
--- a/pytorch_pfn_extras/config_types.py
+++ b/pytorch_pfn_extras/config_types.py
@@ -4,12 +4,12 @@ from .config import Config
 
 def optuna_types(trial):
     types = {
-        "suggest_categorical": trial.suggest_categorical,
-        "suggest_discrete_uniform": trial.suggest_discrete_uniform,
-        "suggest_float": trial.suggest_float,
-        "suggest_int": trial.suggest_int,
-        "suggest_loguniform": trial.suggest_loguniform,
-        "suggest_uniform": trial.suggest_uniform,
+        "optuna_suggest_categorical": trial.suggest_categorical,
+        "optuna_suggest_discrete_uniform": trial.suggest_discrete_uniform,
+        "optuna_suggest_float": trial.suggest_float,
+        "optuna_suggest_int": trial.suggest_int,
+        "optuna_suggest_loguniform": trial.suggest_loguniform,
+        "optuna_suggest_uniform": trial.suggest_uniform,
     }
     return types
 
--- a/pytorch_pfn_extras/config_types.py
+++ b/pytorch_pfn_extras/config_types.py
@@ -1,0 +1,23 @@
+import warnings
+
+from .config import Config
+
+def optuna_types(trial):
+    types = {
+        "suggest_categorical": trial.suggest_categorical,
+        "suggest_discrete_uniform": trial.suggest_discrete_uniform,
+        "suggest_float": trial.suggest_float,
+        "suggest_int": trial.suggest_int,
+        "suggest_loguniform": trial.suggest_loguniform,
+        "suggest_uniform": trial.suggest_uniform,
+    }
+    return types
+
+def load_path_with_optuna_types(path, trial, loader=None, types=None):
+    if types is None:
+        types = {}
+    for key, value in optuna_types(trial).items():
+        if key in types:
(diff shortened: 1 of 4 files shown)


kmaehashi:
@Hakuyume @yuyu2172 Could you take a look on this?

tests/pytorch_pfn_extras_tests/test_config_types.py
```diff
@@ -0,0 +1,75 @@
...
+            self.assertEqual(types['suggest_loguniform'], trial.suggest_loguniform)
+            self.assertEqual(types['suggest_uniform'], trial.suggest_uniform)
+            return 0.0
+        self.study.optimize(objective, n_trials=1)
+
+    def test_load_path_with_optuna_types(self):
+        low = 0
+        high = 8
+        def objective(trial):
+            with tempfile.TemporaryDirectory() as temp0:
```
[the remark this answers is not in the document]
dhgrs: It makes sense. I'm going to do that.

pytorch_pfn_extras/config_types.py
```diff
@@ -0,0 +1,23 @@
+import warnings
+
+from .config import Config
+
+def optuna_types(trial):
+    types = {
+        "suggest_categorical": trial.suggest_categorical,
```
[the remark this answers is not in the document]
dhgrs: I don't have any preference. I agree that `optuna_suggest_*` causes fewer name conflicts so that I'm going to rename them.

setup.py
```diff
@@ -16,7 +16,7 @@
     license='MIT License',
     install_requires=['numpy', 'torch'],
     extras_require={
-        'test': ['pytest'],
+        'test': ['pytest', 'optuna'],
```
dhgrs, line 19: Could you let me know how to add dependency for tests? I thought just adding libraries into `setup.py`, but tests failed.
Should I add them into `.flexci/test.sh` instead of `setup.py`, right?
https://github.com/dhgrs/pytorch-pfn-extras/blob/2dfbffa6b1cc1f14c2712d0c884b05fa8369d8d4/.flexci/test.sh#L5

Hakuyume: @kmaehashi

kmaehashi: Currently test dependencies must be listed in `test.sh`, but I think it's better to list them in `setup.py` instead. You can either do thatin this PR, or just add `optuna` to `test.sh` so that we will work on it in another PR.

dhgrs: Thank you for your answer. Then I'm going to add a commit like
- Add `optuna` to `flexci/test.sh`
- Remove `optuna` in `setup.py` to avoid a conflict in another PR.

kmaehashi:
The test seems failing. (note that all warnings are treated as error via filterwarnings in `pytest.ini`)

dhgrs:
Now, the code warns when user uses pre-defined keys for Optuna like `optuna_suggest_categorical`. I'll add lines to deal with the warnings. I hope I can learn how to deal with warnings.
https://github.com/pfnet/pytorch-pfn-extras/blob/master/tests/pytorch_pfn_extras_tests/training_tests/extensions_tests/test_plot_report.py

dhgrs:
I checked that the latest commit catches all warnings in my local env. Can you test it?

kmaehashi:
@dhgrs Could you check flake8 errors? LGTM otherwise :+1:
https://ci.preferred.jp/pytorch-pfn-extras/64793/#L513

dhgrs:
Fixed it. I checked that flake8 succeeds locally. Although I'm not sure the version of flake8 for CI, it might pass. Could you run the test again?

PR #78, opened 2020-10-11
Add new transforms to `transform_batch`

emcastillo:


emcastillo:
@Hakuyume PTAL 🙂

pytorch_pfn_extras/dataset/tabular/_transform.py
```diff
@@ -2,7 +2,7 @@
 from pytorch_pfn_extras.dataset.tabular import _utils
 
 
-class _Transform(tabular_dataset.TabularDataset):
+class _TransformBase:
```
Hakuyume, line 5: How about making `_TransformBase` as inheritance of `TabularDataset`? It can avoid multiple inheritance in `_Transform` and `_TransformBatch`.

emcastillo: Yes, this makes more sense, thanks!

Hakuyume (approved):
LGTM

emcastillo:
@Hakuyume can you merge this 😇
thanks!!

PR #79, opened 2020-10-15
Add metrics to evaluator

emcastillo:
Add some hooks to calculate metrics easily in the evaluation, this avoids having to override evaluators with all the boilerplate code


asi1024:
LGTM!

PR #80, opened 2020-10-23
Add an alternative to torch.nn.parallel.DistributedDataParallel

HiroakiMikami:
Add a module for data parallelism to handle `torch.util.checkpoint` and dynamic computational graph correctly.

### DIFF
--- a/tests/pytorch_pfn_extras_tests/nn_tests/parallel_tests/test_distributed.py
+++ b/tests/pytorch_pfn_extras_tests/nn_tests/parallel_tests/test_distributed.py
@@ -1,0 +1,259 @@
+import numpy as np
+import os
+import torch
+import tempfile
+import pytest
+from torch import multiprocessing as mp
+from torch import distributed as dist
+from torch import nn
+from torch.utils.checkpoint import checkpoint
+from pytorch_pfn_extras.nn.parallel import DistributedDataParallel
+
+
+torch_version = tuple([int(x) for x in torch.__version__.split(".")])
+context = mp.get_context("spawn")
+
+
+class Hooks:
+    @staticmethod
+    def _to_zero_hook(module):
+        for param in module.parameters():
+            param.grad = torch.zeros_like(param.data)
+
+
+class Steps:
+    @staticmethod
+    def _step(module, input):
+        output = module(input)
+        output.backward()
+        return output
+
+    @staticmethod
+    def _step_with_no_sync(module, input):
+        with module.no_sync():
+            output = module(input)
+            output.backward()
+        return output
+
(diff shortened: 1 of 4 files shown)


emcastillo (approved):
This looks awesome

tests/pytorch_pfn_extras_tests/nn_tests/parallel_tests/test_distributed.py
```diff
@@ -0,0 +1,259 @@
...
+    def forward(self, x):
+        y = self.l0(x)
+        y = checkpoint(self.l1, y)
+        y = checkpoint(self.l2, y)
+        return y
+
+
+def _run(init_file, input, module, rank, args, step, device_type):
+    dist.init_process_group(backend="gloo", init_method=f"file://{init_file}",
+                            world_size=2, rank=rank)
```
emcastillo, line 86: We need to ensure that there are 2 gpus in the tests, sorry I overlooked this.

HiroakiMikami: I think these tests require only 1 gpu.
It's true that I used 2 processes for these tests. But both processes use same GPUs: https://github.com/pfnet/pytorch-pfn-extras/pull/80/files/7386f3a7bdeab2d5b5676172d1dda0e54cd8db39#diff-336610f8599a77e118a5a09f91cbefc98c62f66e9928cd98927abebd8f4b8694R90

emcastillo: ah yes, we are not setting a local device per process :), but I think its better if we do so

HiroakiMikami: I set the local device for each process to `cuda:0` in line 90 of `test_distributed.py`:

```python
    elif device_type == "cuda":
        device = torch.device(f"{device_type}:0")
```

emcastillo: Can we use two gpus instead? we have modified the CI settings to allocate 2 gpus, so it should test this under similar to real-world conditions :)

HiroakiMikami: ok, I'm changing the test script now.

HiroakiMikami: I modified the test cases for using 2 GPUs.

PR #83, opened 2020-10-27
Rename ExtensionManager.{_models,_optimizers} to {models,optimizers} respectively so that extensions can easily access them

belltailjp:
`ExtensionManager` receives models and optimizers and stores them in its internal (in terms of pep8 standard) members.
We often want to access to those model and/or optimizer from an extension. In order to do so in the current interface, especially for a function-based extension created using the `make_extension` decorator, it has to access to the internal members ((1-1) in the example below) or something a little hacky ((1-2), (1-3)).

Therefore I think that it might be handy if such members are treated as "public" interfaces. This PR is to introduce this change by simply renaming {_models,_optimizers} to {models,optimizers}.

For example:

```python
# (1-1) Current interface
@ppe.training.extension.make_extension
def print_conv1_std(manager):
    v = manager._models['main'].conv1.weight.std().item()    # <-- access to the model through _models
    ppe.reporting.report({'conv1_std': v})


# (1-2) Current interface; if the user hesitates to access to underscored methods (nor defining a class)
@ppe.training.extension.make_extension
def print_conv1_std(model, manager):
    v = model.conv1.weight.std().item()
    ppe.reporting.report({'conv1_std': v})

manager.extend(partial(print_conv1_std, model))    # <--- Use partial when adding to the manager


# (1-3) Current interface; dynamic function definition

model = ...

@ppe.training.extension.make_extension
def print_conv1_std(manager):
    global model    # <---- necessary depending on the scope
    v = model.conv1.weight.std().item()    # <-- access to the model defined in the wider scope
    ppe.reporting.report({'conv1_std': v})

# (2) In this PR
@ppe.training.extension.make_extension
def print_conv1_std(manager):
    v = manager.models['main'].conv1.weight.std().item()
    ppe.reporting.report({'conv1_std': v})
```

### DIFF
--- a/pytorch_pfn_extras/training/extensions/fail_on_non_number.py
+++ b/pytorch_pfn_extras/training/extensions/fail_on_non_number.py
@@ -13,7 +13,7 @@ class FailOnNonNumber(extension.Extension):
     """
 
     def __call__(self, manager):
-        for name, model in manager._models.items():
+        for name, model in manager.models.items():
             for param in model.parameters():
                 if (not torch.isfinite(param).all()
                         or (param.grad is not None
--- a/pytorch_pfn_extras/training/manager.py
+++ b/pytorch_pfn_extras/training/manager.py
@@ -80,17 +80,17 @@ class _BaseExtensionsManager:
             if not isinstance(models, torch.nn.Module):
                 raise ValueError(
                     'model must be an instance of dict or toch.nn.Module')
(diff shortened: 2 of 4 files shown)


emcastillo:
How about using a property? 🙂

belltailjp:
That makes sense! I have updated the PR.

emcastillo (approved):

PR #85, opened 2020-10-28
Change flexci limits

emcastillo:
For #80 tests to work properly


emcastillo:
] API failed: /a/set_job: memory per CPU limit is exceeded: 8.000000 > 6.500000

kmaehashi:
It's a restriction of GCP... Could you add one more CPU core?

PR #86, opened 2020-11-12
Name inputs and remove unused inputs from testcase

okdshin:
When not given input_names, inputs are named by torch.onnx. The naming rule is not controllable and seems inconsistent, so we should name inputs by ourself.
With our naming, we can detect unused inputs and remove them from testcases.

### DIFF
--- a/pytorch_pfn_extras/onnx/export_testcase.py
+++ b/pytorch_pfn_extras/onnx/export_testcase.py
@@ -154,10 +154,19 @@ def export_testcase(
     """
 
     os.makedirs(out_dir, exist_ok=True)
+    input_names = kwargs.pop(
+        'input_names',
+        ['input_{}'.format(i) for i in range(len(args))])
 
     onnx_graph, outs = _export(
         model, args, strip_large_tensor_data, large_tensor_threshold,
-        **kwargs)
+        input_names=input_names, **kwargs)
+
+    # Remove unused inputs
+    used_input_index_list = [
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -154,10 +154,19 @@ def export_testcase(
...
     onnx_graph, outs = _export(
         model, args, strip_large_tensor_data, large_tensor_threshold,
-        **kwargs)
+        input_names=input_names, **kwargs)
+
+    # Remove unused inputs
+    used_input_index_list = [
+        input_names.index(input.name) for input in onnx_graph.graph.input]
+    input_names = [input_names[i] for i in used_input_index_list]
+    args = [args[i] for i in used_input_index_list]
```
disktnk, line 169: is it better to check `assert len(args) == len(input_names)` before?

okdshin: I added here https://github.com/pfnet/pytorch-pfn-extras/pull/86/files#diff-4b250f25742298aa0ef7ea68986407898b68b8d35420d0384804d1e29d4f3671R160.

take-cheeze:
Seems few tests is failing: https://ci.preferred.jp/pytorch-pfn-extras/64749/#L383

take-cheeze (approved):
LGTM 👍

PR #88, opened 2020-11-27
Add `LRScheduler` extension

emcastillo:
Needed for an easily interoperability of pytorch lr schedulers and ppe. This will automatically call step and take snapshots when required

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -8,6 +8,7 @@ Currently the following extensions are available:
 
 + `Evaluator`
 + `LogReport`
++ `LRScheduler`
 + `MicroAverage`
 + `PrintReport`
 + `ProgressBar`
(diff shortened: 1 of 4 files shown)


kmaehashi:
LGTM!
/test

kmaehashi:
Several lr_schedulers requires argument to `step` method. How about allowing users to customize it?

```py
def _get_value_from_log_report(manager, key):
    # Find and return the latest reported "key" from LogReport

def _default_stepper(manager, scheduler):
    if isinstance(scheduler, ReduceLROnPlateau):
        LRScheduler.step_by_value()()
    scheduler.step()

class LRScheduler(Extension):
    def __init__(self, scheduler, *, stepper=_default_stepper):
        ...

    def __call__(self, manager):
        self.stepper(manager, self.scheduler)

    @staticmethod
    def step_by_value(key='main/loss'):
        def _stepper(manager, scheduler):
            scheduler.step(_get_key_from_log_report(manager, key))
        return _stepper
```

So users can write:

```
# simply call step
manager.extend(LRScheduler(StepLR(...)))

# step by the latest loss (main/loss)
manager.extend(LRScheduler(ReduceLROnPlateau(...)))

# step by the latest loss (custom report key)
manager.extend(LRScheduler(ReduceLROnPlateau(...), stepper=LRScheduler.step_by_value('my/loss')))
```

PR #94, opened 2020-12-07
Add trainer

emcastillo:
Add a customizable training framework that can be expanded to deal with multiple devices.


kmaehashi:
We'll work on this feature in `trainer` branch.

PR #95, opened 2020-12-08
Fix torch backend

emcastillo:


pytorch_pfn_extras/training/trainer.py
```diff
@@ -76,6 +78,9 @@ def takes_ckpt_this_iter(self, idx):
     def get_to_report_outputs(self):
         return self.to_report_outputs
 
+    def get_backward_variable_names(self):
```
kmaehashi, line 81: How about passing this to train_step, instead of accessing `get_backward_variable_names` from backends?

PR #97, opened 2020-12-14
Fix snapshot not closing file

emcastillo:
Leaving files unclosed in the autoload causes a warning that triggers and exception in the latest pytest version

### DIFF
--- a/pytorch_pfn_extras/training/extensions/_snapshot.py
+++ b/pytorch_pfn_extras/training/extensions/_snapshot.py
@@ -367,6 +367,7 @@ class _Snapshot(extension.Extension):
                         target[k].load_state_dict(state[k], **kwargs)
                 else:
                     target.load_state_dict(state, **kwargs)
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/training/extensions/_snapshot.py
```diff
@@ -367,6 +367,7 @@ def initialize(self, manager):
                         target[k].load_state_dict(state[k], **kwargs)
                 else:
                     target.load_state_dict(state, **kwargs)
+                writer.fs.close(snapshot_file)
```
kmaehashi, line 370: In PFIO, this is an interface to shutdown the filesystem itself. We should call close on file instead.

```suggestion
                snapshot_file.close()
```

Note: HdfsFile object provides `close` method
https://arrow.apache.org/docs/python/generated/pyarrow.HdfsFile.html#pyarrow.HdfsFile

kmaehashi (changes requested):

PR #102, opened 2020-12-22
Add add_value_names option to _model_to_graph

take-cheeze:
By default value names exported by torch.onnx is integer unique id of value:
https://github.com/pytorch/pytorch/blob/42c78ed74525f03f6bb43110784f07c5a6ef1bef/torch/csrc/jit/ir/ir.h#L190-L195

But integer is not human recognizable format so using same mechanism as setting names of inputs/outputs:
https://github.com/pytorch/pytorch/blob/e1c1a7e96421735f1c6397de1e32e9986bd839d0/torch/onnx/utils.py#L737-L749

the value names would be `v{unique_id}_{node op_type}` with this PR.

And the names of value would be C identifier compatible. Ref: https://github.com/onnx/onnx/blob/master/docs/IR.md#names-within-a-graph

### DIFF
--- a/pytorch_pfn_extras/onnx/export_testcase.py
+++ b/pytorch_pfn_extras/onnx/export_testcase.py
@@ -43,9 +43,6 @@ def _model_to_graph_with_value_names(*args, add_value_names=True, **kwargs):
     return g, p, o
 
 
-torch.onnx.utils._model_to_graph = _model_to_graph_with_value_names
-
-
 def _export_meta(model, out_dir, strip_large_tensor_data):
     ret = {
         'generated_at': datetime.datetime.now().isoformat(),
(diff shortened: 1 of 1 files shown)


pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -22,6 +24,28 @@
...
+            new_name = 'v{}_{}'.format(old_name, n.kind().split('::')[-1])
+            v.setDebugName(new_name)
+            if old_name in p:
+                i = p[old_name]
+                del p[old_name]
+                p[new_name] = i
+    return g, p, o
+
+
+torch.onnx.utils._model_to_graph = _model_to_graph_with_value_names
```
take-cheeze, line 46: memo: replacing only in ppe.onnx.export would be more cleaner (thank you @kmaehashi san)

pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -92,6 +92,8 @@ def _export_util(model, args, f, **kwargs):
             operator_export_type = OperatorExportTypes.ONNX
 
     old_model_to_graph = torch.onnx.utils._model_to_graph
+    # TODO(ecastill) _model_to_graph shouldn't be direclty overriden
+    # This is a temporal workaround until a fix is introduced in PyTorch.
```
take-cheeze, line 96: 👍

PR #115, opened 2021-02-08
Fix autoload and transform models conflict

emcastillo:
When loading a model wrapped in DDP or already transformed.
The logic for transforming the model should be the opposite of what we used to do.
Otherwise, autoload will require to modify the script when we want to resume the training.

With this PR, `autoload_transform_models` now has no sense.


kmaehashi:
LGTM!

PR #116, opened 2021-02-08
Add LRScheduler extension

emcastillo:


emcastillo:
Needs to clean f-strings

PR #117, opened 2021-02-08
Add BN aware checkpointing

emcastillo:
`torch.utils.checkpoint` is not dealing correctly with batch normalization updating the internal stats twice as two forward passes are done. 
This PR is a solution implemented by @tkerola (all credit is his!) that reduces the momentum so both updates result in the same end value.

### DIFF
--- a/pytorch_pfn_extras/utils/checkpoint.py
+++ b/pytorch_pfn_extras/utils/checkpoint.py
@@ -10,8 +10,8 @@ class _CheckpointFunction(torch.utils.checkpoint.CheckpointFunction):
     momentum is updated twice, while we only need one update to ensure
     correctness.
     Using `ppe.utils.checkpointing.checkpoint` as a drop-in replacement
-    can help to lead with incorrect values in the backward pass generated
-    by this issue.
+    can help deal with incorrect values in the BatchNormalization
+    persistent parameters.
     """
     @staticmethod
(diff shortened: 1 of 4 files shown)


pytorch_pfn_extras/utils/checkpoint.py
```diff
@@ -0,0 +1,42 @@
...
+class _CheckpointFunction(torch.utils.checkpoint.CheckpointFunction):
+    """Checkpoint a model or part of the model with BN support.
+
+    Refer to https://pytorch.org/docs/stable/checkpoint.html
+    for detailed information.
+    When using checkpointing in model using BatchNormalization, the
+    momentum is updated twice, while we only need one update to ensure
+    correctness.
+    Using `ppe.utils.checkpointing.checkpoint` as a drop-in replacement
+    can help to lead with incorrect values in the backward pass generated
```
tkerola, line 13: Actually, this PR doesn't change the values in the backward pass. It just makes sure that the persistent avg mean/var parameters are updated correctly (which will be used during test time). Gradients will be the same with and without this PR.

tkerola: Perhaps the doc should be something like
```Using `ppe.utils.checkpointing.checkpoint` as a drop-in replacement
    can help deal with incorrect values in the BatchNormalization persistent parameters.
```

kmaehashi (approved):

PR #118, opened 2021-02-15
Add `append` and `format` option in `LogReport`

asi1024:
TODO: add tests

### DIFF
--- a/pytorch_pfn_extras/training/extensions/log_report.py
+++ b/pytorch_pfn_extras/training/extensions/log_report.py
@@ -106,7 +106,7 @@ keys=None, trigger=(1, 'epoch'), postprocess=None, filename='log', writer=None)
         del log_name  # avoid accidental use
         self._log_name = filename
 
-        if format is None:
+        if format is None and filename is not None:
             if filename.endswith('.jsonl'):
                 format = 'json-lines'
             elif filename.endswith('.yaml'):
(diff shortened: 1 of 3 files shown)


pytorch_pfn_extras/training/extensions/log_report.py
```diff
@@ -12,8 +12,22 @@
...
+class LogWriterSaveFunc:
+
+    def __init__(self, format):
+        self._format = format
+
+    def __call__(self, target, file_o):
+        if self._format == 'json':
+            log = json.dumps(target, indent=4)
+        elif self._format == 'json-lines':
+            log = '\n'.join([json.dumps(x) for x in target])
```
kmaehashi, line 24: Could you support appending the result so that only diff is written when in append=True mode?

tests/pytorch_pfn_extras_tests/training_tests/extensions_tests/test_print_report.py
```diff
@@ -3,18 +3,29 @@
...
     iters_per_epoch = 5
     manager = ppe.training.ExtensionsManager(
         {}, {}, max_epochs, iters_per_epoch=iters_per_epoch)
 
     out = io.StringIO()
-    log_report = ppe.training.extensions.LogReport()
+    log_report = extensions.LogReport(format=format, append=append)
     manager.extend(log_report)
-    extension = ppe.training.extensions.PrintReport(out=out)
+    extension = extensions.PrintReport(out=out)
```
emcastillo, line 28: I think this is not testing that the file is being written.
but it wasn't doing it before either. So we should do it in a different PR.

emcastillo (approved):

emcastillo:
TODO: Add a proper `LogReport` test that writes the file and parses it back. 
There is currently no `LogReport` test, so I will be adding it in a different PR

emcastillo:
CI fails are real :(

PR #119, opened 2021-02-15
Fix CI error due to new numpy

emcastillo:
Temporary skip numpy 1.20 deprecation warning as the most stable cupy version does not have guaranteed support for it.


emcastillo:
Dup #120

PR #124, opened 2021-02-18
Fix summary snapshot restore

### LINKED ISSUE #122: Snapshot extensions can't deal with `reporting.Summary` when the tensor is on GPU
As described here, `Summay.add` can deal with a tensor on GPU.
https://github.com/pfnet/pytorch-pfn-extras/blob/5024cf746846dd41a010eedb287925ee09b17e43/pytorch_pfn_extras/reporting.py#L261-L271

When calling `add` after snapshot extensions load summary from their snapshot, it causes an error because of multiple devices.

```python
import torch

SNAPSHOT_FILENAME = "snapshot"

tensor = torch.zeros(0, device="cuda")
summary = ppe.reporting.Summary()
summary.add(tensor)
torch.save(summary.state_dict(), SNAPSHOT_FILENAME)

manager = ppe.training.ExtensionsManager(
    {}, None, None, iters_per_epoch=1, out_dir="./"
)
snapshot = ppe.training.extensions.snapshot_object(
    summary,
    SNAPSHOT_FILENAME,
    autoload=True,
)

snapshot.initialize(manager)
summary.add(tensor)
```

Output:
```
  File "/tmp/a.py", line 21, in <module>
    summary.add(tensor)
  File "/usr/local/lib/python3.8/site-packages/pytorch_pfn_extras/reporting.py", line 272, in add
    self._x += weight * value
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

emcastillo:
Closes #122

We know the stats are scalars, so we will be saving and restoring them using their primitive python dtypes.

### DIFF
--- a/pytorch_pfn_extras/reporting.py
+++ b/pytorch_pfn_extras/reporting.py
@@ -297,15 +297,21 @@ class Summary:
     def state_dict(self):
         state = {}
         try:
-            state = {'_x': self._x, '_x2': self._x2, '_n': self._n}
+            # Save the stats as python scalars in order to avoid
+            # different device errors when loading them back
+            state = {'_x': float(self._x),
+                     '_x2': float(self._x2),
+                     '_n': int(self._n)}
         except KeyError:
             warnings.warn('The previous statistics are not saved.')
         return state
 
     def load_state_dict(self, to_load):
-        self._x = _nograd(to_load['_x'])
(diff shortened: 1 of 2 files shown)


kmaehashi (approved):
Sorry for the delay, LGTM!

PR #126, opened 2021-03-08
Fix long keys in from_data

Hakuyume:
Currently, `tabular.from_data` treats `key_a` as a tuple of characters. This must be a bug.


emcastillo (approved):
Good catch!!! 🙇‍♂️

kmaehashi:
@Hakuyume Could you check coding style errors? :pray:

emcastillo:
I will fix them since they are only style issues :)

Hakuyume:
> I will fix them since they are only style issues :)

Thanks!!

PR #135, opened 2021-03-15
Fix get_examples of _Transform/_TransformBatch

Hakuyume:
When `get_examples` is called with discontinuous `key_indices`,  `_Transform/_TransformBatch` creates wrong mapping between the outputs of transform functions and the outputs of  `get_examples`.


emcastillo (approved):
LGTM!

kmaehashi:
Test passed.

PR #125, opened 2021-03-08
Add tests for PyTorch 1.8

kmaehashi:
- [x] Create FlexCI project `pytorch-pfn-extras.torch18-linux` and `pytorch-pfn-extras.torch18-win`.
- [x] Add a webhook for these projects.
- [ ] Trigger a docker image build (`prep` project) manually and wait until the docker image is built. https://ci.preferred.jp/pytorch-pfn-extras.prep-linux/69916/
- [ ] Test this PR.

PR #139, opened 2021-04-01
Remove old dataloader

emcastillo:
Now that pytorch DL supports `persistent_workers` since 1.7, and we are removing support for ppe, we don't need to keep around this DL.


kmaehashi (approved):

kmaehashi:
Let's merge #125 first, then test this again and merge.

PR #144, opened 2021-04-19
Add profiler and extensions

emcastillo:
Contributed by @HiroakiMikami 

This adds a profiler used to measure cuda time and cpu time of iterations.

This PR also fixes a bug where the log in append mode wasn't generating correct results

### DIFF
--- a/pytorch_pfn_extras/profiler/__init__.py
+++ b/pytorch_pfn_extras/profiler/__init__.py
@@ -1,2 +1,2 @@
-from pytorch_pfn_extras.profiler import record  # NOQA
-from pytorch_pfn_extras.profiler import time_summary  # NOQA
+from pytorch_pfn_extras.profiler.record import record  # NOQA
+from pytorch_pfn_extras.profiler.time_summary import time_summary  # NOQA
--- a/pytorch_pfn_extras/profiler/__init__.py
+++ b/pytorch_pfn_extras/profiler/__init__.py
@@ -1,0 +1,2 @@
+from pytorch_pfn_extras.profiler import record  # NOQA
+from pytorch_pfn_extras.profiler import time_summary  # NOQA
--- a/pytorch_pfn_extras/profiler/time_summary.py
+++ b/pytorch_pfn_extras/profiler/time_summary.py
@@ -1,0 +1,138 @@
+import atexit
+from contextlib import contextmanager
+import time
+from typing import Tuple
+from threading import Lock, Thread
+from queue import Queue
+import multiprocessing as mp
+import torch
+from pytorch_pfn_extras.reporting import DictSummary
+
+
+Events = Tuple[torch.cuda.Event, torch.cuda.Event]
+
+
+class _CPUWorker(object):
+    def __init__(self, add, max_queue_size: int = 1000):
+        self._add = add
+        self._queue = mp.JoinableQueue(max_queue_size)
+        self._thread = Thread(target=self._worker)
+        self._thread.setDaemon(True)
+        self._thread.start()
+
+    def close(self):
+        self.wait()
+        self._queue.put(None)
+        self._thread.join()
+        self._queue.close()
+        self._queue.join_thread()
+
+    def wait(self):
+        self._queue.join()
+
+    def _worker(self):
+        while True:
+            v = self._queue.get()
+            if v is None:
+                self._queue.task_done()
+                break
+            name, value = v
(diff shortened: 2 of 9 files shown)


pytorch_pfn_extras/profiler/time_summary.py
```diff
@@ -0,0 +1,138 @@
...
+            if v is None:
+                self._queue.task_done()
+                break
+            name, value = v
+            self._add(name, value)
+            self._queue.task_done()
+
+
+class _CUDAWorker(object):
+    def __init__(self, add, max_queue_size: int = 1000):
```
asi1024, line 45: Ditto.

pytorch_pfn_extras/profiler/time_summary.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def _worker(self):
+        while True:
+            v = self._queue.get()
+            if v is None:
+                self._queue.task_done()
+                break
+            name, value = v
+            self._add(name, value)
+            self._queue.task_done()
```
asi1024, line 41: How about using `empty()` to check if all tasks are done instead of putting `None` as a sentinel for the task queue?

```suggestion
        while not self._queue.empty():
            name, value = self._queue.get()
            self._add(name, value)
            self._queue.task_done()
```

pytorch_pfn_extras/profiler/time_summary.py
```diff
@@ -0,0 +1,138 @@
...
+
+class _CUDAWorker(object):
+    def __init__(self, add, max_queue_size: int = 1000):
+        self._add = add
+        self._queue = Queue(max_queue_size)
+        self._thread = Thread(target=self._worker)
+        self._thread.setDaemon(True)
+        self._thread.start()
+        self._event_lock = Lock()
+        self._events = []
```
asi1024, line 52: How about using `Queue` instead?

pytorch_pfn_extras/profiler/__init__.py
```diff
@@ -0,0 +1,2 @@
+from pytorch_pfn_extras.profiler import record  # NOQA
+from pytorch_pfn_extras.profiler import time_summary  # NOQA
```
asi1024, line 2: I think `ppe.profiler.record.record` and `ppe.profiler.time_summary.time_summary` are a little bit long and annoying. How about revealing these attributes here directly?

pytorch_pfn_extras/training/extensions/profile_report.py
```diff
@@ -0,0 +1,102 @@
...
+
+            stats_cpu["epoch"] = manager.epoch
+            stats_cpu["iteration"] = manager.iteration
+            stats_cpu["elapsed_time"] = manager.elapsed_time
+            # Recreate dict to fix order of logs
+            out = {}
+            keys = list(stats_cpu.keys())
+            keys.sort()
+            for key in keys:
+                out[key] = stats_cpu[key]
```
asi1024, line 80: It is equivalent to `out = stats_cpu.copy()`?

asi1024 (commented):
Could you add tests?

pytorch_pfn_extras/training/extensions/profile_report.py
emcastillo: I think the idea here was to enforce the order, `stats_cpu` might show keys in a different order, so dict recreation ensures it

pytorch_pfn_extras/profiler/time_summary.py
emcastillo: unfortunately this will not execute the loop at all, this runs in a thread and it needs to loop always until the centinel is retrieved :(

So when the thread begins, the queue is empty and it never consumes the value.

pytorch_pfn_extras/training/extensions/profile_report.py
asi1024: Python's built-in dictionary does not guarantee the consistency of the order of keys. We should use `collections.OrderedDict()`?

emcastillo: Done! thanks!
add a fix to log reporter save_fn to save `OrderedDicts`

emcastillo:
added test for the extension! PTAL!

asi1024 (approved):

emcastillo:
seems like test_time_summary hangs the CI

PR #153, opened 2021-05-06
Fix minmax value trigger

emcastillo:
This causes errors when using DDP as some of the workers may not have the key in their stats.


emcastillo:
I think it is safe to review this :)

kmaehashi:
Closing as #166 got merged.

PR #158, opened 2021-05-11
Fix to finalize TensorBoardWriter

kmaehashi:
TensorBoard's SummaryWriter must be finalized, but it was not.

This PR walso adds LogReport & TensorBoardWriter use-case test (c.f., #156).

PR #159, opened 2021-05-11
Fix TensorBoard to no-op if TensorBoard is unavailable

### LINKED ISSUE #155: Make TensorBoardWriter no-op if tensorboard is not available

kmaehashi:
Closes #155. ~Depends on #158~ (merged).
I think this is better when reusing experiment codes.


kmaehashi:
Rebased.
/test

kmaehashi:
Hmm seems it hangs at the end of epoch? Let me test locally...

emcastillo:
Depends on #433

kmaehashi:
Thanks! LGTM :+1:

Issue #162, opened 2021-05-17
Introduce type annotations


kmaehashi:
As we assume users inherit some APIs, it's better to have a type annotation.

PR #164, opened 2021-05-18
Use pysen and enable mypy check

asi1024:
Part of #162.

Added mypy test to flexCI configuration and made small fixes to pass the test. I'm planning to send another PR for further type annotations after this PR.


kmaehashi:
Discussions:
- Introduce pysen and follow mypy option generated.
- Do not add type annotation to tests, because it is difficult to annotate parameterize tests and tests using mock, so we won't annotate tests. Increase the coverage of example and add types there to detect typing errors.

asi1024:
Introduced pysen (136414a).
```
Found 1058 errors in 57 files (checked 74 source files)
```

asi1024:
@kmaehashi PTAL!

kmaehashi:
Could you remove package_data from setup.py as well?

asi1024:
@kmaehashi Fixed!

PR #165, opened 2021-05-19
Fix example to init lazy layers after transferring to device

kmaehashi:
It is better for performance.


kmaehashi:
Test passed.

asi1024:
LGTM.

PR #169, opened 2021-05-25
Fix timer function

kmaehashi:
`time.perf_counter` is added in Python 3.3 so we can now assume it's always available.
https://docs.python.org/3.6/library/time.html#time.perf_counter

Also note that `time.clock` is removed in Python 3.8.


kmaehashi:
Test passed.

emcastillo (approved):
😂

PR #170, opened 2021-05-25
Enable flake8 in pysen

asi1024:
~Blocked by #164~

### DIFF
--- a/tests/pytorch_pfn_extras_tests/dataset_tests/tabular_tests/test_slice.py
+++ b/tests/pytorch_pfn_extras_tests/dataset_tests/tabular_tests/test_slice.py
@@ -93,7 +93,7 @@ params = _filter_params(product_dict(
         ],
     )
     + product_dict(
-
+        [{'mode': None}],
         [
             {'keys': None, 'expected_keys': ('a',)},
             {'keys': 0, 'expected_keys': ('a',)},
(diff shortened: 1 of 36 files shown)


tests/pytorch_pfn_extras_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -62,7 +62,7 @@ def test_call():
 def test_savefun_and_writer_exclusive():
     # savefun and writer arguments cannot be specified together.
     def savefun(*args, **kwargs):
-        assert False
+        pytest.fail('never reach')
```
asi1024, line 65: note: flake8 does not allow `assert False` because of B011.
```
B011 Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().
```
flake8 suggests using `raise AssertionError(...)`, but I replaced this line with `pytest.fail` for the readability.

tests/pytorch_pfn_extras_tests/dataset_tests/tabular_tests/test_slice.py
```diff
@@ -91,9 +91,9 @@ def _filter_params(params):
             {'keys': (-1, 'a'), 'expected_keys': ('c', 'a')},
             {'keys': (), 'expected_keys': ()},
         ],
-    ) +
-    product_dict(
-        [{'mode': None}],
```
kmaehashi, line 96: Is this removed as intended?

asi1024: Good catch! Fixed!

kmaehashi (approved):

PR #174, opened 2021-05-28
Fix print in `_PosixFileSystem.rename`

hvy:
Fixes a print message in `_PosixFileSystem.rename` which could be a bit confusing.


kmaehashi (approved):
LGTM!

kmaehashi:
Thanks!

PR #178, opened 2021-05-31
Make _tensor.py typed

asi1024:
Related discussion: https://github.com/python/mypy/issues/1153#issuecomment-558555225


pytorch_pfn_extras/_cupy/cupy.py
```diff
@@ -0,0 +1,2 @@
+class ndarray:
+    pass
```
kmaehashi, line 2: Could you rename this file to `_cupy_stub.py`? I think it's better to avoid filenames that conflicts with the real module name.

pytorch_pfn_extras/_cupy/__init__.py
```diff
@@ -0,0 +1,17 @@
+try:
+    import cupy  # NOQA
+    _cupy_import_error = None
+except Exception as e:
+    from pytorch_pfn_extras._cupy import cupy
+    cupy = None  # NOQA
```
kmaehashi, line 6:
```suggestion
    from pytorch_pfn_extras._cupy._stub_cupy import cupy
```

kmaehashi (changes requested):

asi1024:
@kmaehashi PTAL!

kmaehashi (approved):
LGTM! :+1:

Issue #168, opened 2021-05-24
Error writing snapshot backup file


niboshi:
https://github.com/pfnet/pytorch-pfn-extras/blob/1b265e7de61f94cbb0910dbd60aa86a719ebdb7d/pytorch_pfn_extras/writing.py#L216

This line caused an error.
It seems `.bak` file already existed at the time of this call. I think such a situation can easily happen if the previous training session was terminated before [removing it](https://github.com/pfnet/pytorch-pfn-extras/blob/1b265e7de61f94cbb0910dbd60aa86a719ebdb7d/pytorch_pfn_extras/writing.py#L221).

By the way, it also failed to load the snapshot of the previous session and started over from the epoch 1 (see the timestamps of the files listed below).
I don't know whether or not it's related to this issue.

```
epoch       iteration   elapsed_time  lr          train/loss  val/loss    val/top1    val/top5                                                                                                                                       
1           58          17.791        0.1         8.19181     31.9144     0           0.0119048                                                                                                                                      
2           116         40.637        0.1         7.68947     7.59687     0.00892857  0.00892857                                                                                                                                     
Traceback (most recent call last):                                                                                                                                                                                                   
  File "/usr/local/lib/python3.8/runpy.py", line 194, in _run_module_as_main                                                                                                                                                         
    return _run_code(code, main_globals, None,                                                                                                                                                                                       
  File "/usr/local/lib/python3.8/runpy.py", line 87, in _run_code                                                                                                                                                                    
    exec(code, run_globals)                                                                                                                                                                                                          
  File "mywork/__main__.py", line 40, in <module>                                      
    main()                                                                                                                                                                                                                           
  File "mywork/__main__.py", line 34, in main                                          
    args.func(args)                                                                                                                                                                                                                  
  File "mywork/train.py", line 422, in main                                            
    run_train(manager, model, device, train_loader, optimizer,                                                                                                                                                                       
  File "mywork/train.py", line 153, in run_train                                       
    ppe.reporting.report({'train/loss': loss.item()})                                                                                                                                                                                
  File "/usr/local/lib/python3.8/contextlib.py", line 120, in __exit__                                                                                                                                                               
    next(self.gen)                                                                                                                                                                                                                   
  File "/home/nishino/.local/lib/python3.8/site-packages/pytorch_pfn_extras/training/manager.py", line 430, in run_iteration                                                                                                         
    self.run_extensions()                                                                                                                                                                                                            
[...]

PR #185, opened 2021-06-07
Improve handling of backup files

emcastillo:
This is to partly fix the issue in #168 
Some unexpected process termination can leave a '.bak' file undelete, so next executions will error.

### DIFF
--- a/pytorch_pfn_extras/writing.py
+++ b/pytorch_pfn_extras/writing.py
@@ -220,7 +220,7 @@ class Writer:
             if make_backup:
                 bak = '{}.bak'.format(dest)
                 # Check if another backup file exists
-                # due to some unexpected termination of an older
+                # due to some unexpected termination of an earlier
                 # process
                 if self.fs.exists(bak):
                     self.fs.remove(bak)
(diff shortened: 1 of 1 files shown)


pytorch_pfn_extras/writing.py
```diff
@@ -217,6 +217,11 @@ def save(self, filename, out_dir, target,
             make_backup = self.fs.exists(dest)
             if make_backup:
                 bak = '{}.bak'.format(dest)
+                # Check if another backup file exists
+                # due to some unexpected termination of an older
```
kmaehashi, line 221: nit :smile:
```suggestion
                # due to some unexpected termination of an earlier
```

kmaehashi (approved):

kmaehashi (approved):

kmaehashi:
LGTM. This PR also changes the order of handling snapshot files.

Previously, the logic was:

1. Rename dest to dest.bak
2. Create tmp_dest
3. Rename tmp_dest to dest

However, step 2 often takes time so there was a high possibility of being preempted (killed) during that process. This PR changes the logic to:

1. Create tmp_dest
2. Rename dest to dest.bak
3. Rename tmp_dest to dest

so that `dest` is kept even if the process is killed during the file creation. Note that it is still possible that the process is killed between 2 and 3 but we need this non-atomic operation to support various file systems e.g. HDFS.

PR #186, opened 2021-06-08
Make autoload fault tolerant

emcastillo:
Needs discussion

In some situations, temporal snapshot files might be found instead of final ones,
If only such files exist, is it reasonable to resume from them? In scenarios when training is automatically resumed after a fault
this can be a good idea to not waste computing resources.

Should we control this behavior with a keyword argument for snapshot? (I think this is probably the best option)

### DIFF
--- a/pytorch_pfn_extras/training/extensions/_snapshot.py
+++ b/pytorch_pfn_extras/training/extensions/_snapshot.py
@@ -1,9 +1,15 @@
 import os
+import warnings
 
 import torch
 import torch.distributed
 
 from pytorch_pfn_extras.training import extension
+
+
+def _match_files(prefix, suffix, path, fs):
+    return [file for file in fs.list(path)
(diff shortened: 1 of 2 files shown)


dhgrs:
Please let me know, why is this PR closed? Are you working on it in other issues or PRs?

emcastillo:
Sorry, we found a better way to address this in https://github.com/pfnet/pytorch-pfn-extras/pull/185 :)
Can you test master branch now?

dhgrs:
I see, I didn't notice that #185 also fixes this issue. I'll check it.

dhgrs:
I checked the PR. The PR doesn't change the way to find snapshots but changes the frequency where only temporary files exist. Now it very rarely happens to miss existing tmp snapshots. Then, what I can do to check more carefully is just using it in my workloads and making sure that the issue doesn't happen, right?

PR #187, opened 2021-06-08
Fix LazyModules load_hook

### LINKED ISSUE #167: `autoload=True` not working with models in the gpu
If we create the snapshot extension with `autoload=True` the model will not correctly load its state.

`autoload=true` loads the state in the CPU. but it is not executed until the first iteration and it will overwrite the device of the model.  This requires us to call `start_extensions` manually and then do the device move for it to work

```python
# model parameters will be moved to cpu at the beginning of the first iteration due autoload being executed there
...
manager.extend(extensions.snapshot(autoload=True)
# move the model, but this will be overwriten later
model.cuda()

for batch in train_loader:
    with manager.run_iteration():   # Snapshot load happens the first time this is executed
        model(batch.cuda())   #  Error! weights are in the cpu again
```

emcastillo:
Fixes #167 

We should be more careful and avoid weird cases of replacing existing parameters with non-existing ones, so we avoid recreating model parameters in the `load_state` dict which can lead to the new parameters not being tracked by the optimizer.

Torch lazy modules forbid mixing non-initialized params with initialized ones.

### DIFF
--- a/pytorch_pfn_extras/nn/modules/lazy.py
+++ b/pytorch_pfn_extras/nn/modules/lazy.py
@@ -93,20 +93,8 @@ class LazyInitializationMixin:
         for the details of the hook specification.
         """
         for name in self.lazy_buffer_names:
-            # Avoid shape mismatch error when loading an initialized buffer
-            # onto an uninitialized module instance.
             key = prefix + name
-            if key in state_dict:
-                # The model was serialized after initialization.
-                if getattr(self, name).shape == (0,):
-                    raise RuntimeError(
-                        'Can\'t load an already initialized '
-                        'buffer into a non initialized one')
-            else:
-                if getattr(self, name).shape != (0,):
-                    raise RuntimeError(
-                        'Can\'t replace a already initialized '
-                        'buffer with a non initialized one')
+            if key not in state_dict:
                 buffer = torch.Tensor([])
                 state_dict[key] = buffer
 
--- a/pytorch_pfn_extras/nn/modules/lazy.py
+++ b/pytorch_pfn_extras/nn/modules/lazy.py
@@ -81,45 +81,28 @@ class LazyInitializationMixin:
             missing_keys, unexpected_keys, error_msgs):
         """load_state_dict pre-hook function for lazy buffers and parameters.
 
-        The purpose of this hook is to adjust the current state and/or
-        ``state_dict`` being loaded so that a module instance serialized in
-        both un/initialized state can be deserialized onto both un/initialized
-        module instance.
+        The purpose of this hook is to check the current state and/or
+        ``state_dict`` being loaded and ensure that both are states
+        are properly initialized.
 
         See comment in ``torch.nn.Module._register_load_state_dict_pre_hook``
         for the details of the hook specification.
         """
-        for name in self.lazy_buffer_names:
-            key = prefix + name
-            module_initialized = getattr(self, name).shape != (0,)
-            state_initialized = (
-                 key in state_dict and (state_dict[key].shape != (0,)))
-            if module_initialized != state_initialized:
-                raise RuntimeError(
-                    'Can\'t load initialized buffers in non-initialized '
-                    'modules and vice versa')
-            else:
-                # For compatibility with older snapshots
-                if key not in state_dict and not module_initialized:
-                    buffer = torch.Tensor([])
-                    state_dict[key] = buffer
-
         for name in self.lazy_parameter_names:
-            # The parameter may not exist in the loaded ``state_dict`` if the
+            # The parameter does not exist in the loaded ``state_dict`` if the
             # original module was serialized before initializing lazy
             # parameters (see comments of ``state_dict``).
             key = prefix + name
             module_initialized = not isinstance(
                 getattr(self, name), UninitializedParameter)
             state_initialized = key in state_dict
-            if module_initialized != state_initialized:
+            if module_initialized and not state_initialized:
                 raise RuntimeError(
-                    'Can\'t load initialized parameters in non-initialized '
-                    'modules and vice versa')
-            else:
-                if key not in state_dict and not module_initialized:
-                    param = UninitializedParameter()
-                    state_dict[key] = param
+                    'Can\'t load uninitialized parameters in already '
+                    'initialized modules')
+            elif key not in state_dict and not module_initialized:
+                param = UninitializedParameter()
+                state_dict[key] = param
 
 
 class UninitializedParameter(torch.nn.Parameter):
--- a/pytorch_pfn_extras/nn/modules/lazy.py
+++ b/pytorch_pfn_extras/nn/modules/lazy.py
@@ -91,14 +91,13 @@ class LazyInitializationMixin:
         for name in self.lazy_buffer_names:
             key = prefix + name
             module_initialized = getattr(self, name).shape != (0,)
-            state_initialized = (
-                 key in state_dict and (state_dict[key].shape != (0,)))
+            state_initialized = state_dict[key].shape != (0,)
             if module_initialized and not state_initialized:
                 raise RuntimeError(
                     'Can\'t load non-initialized buffers in already '
                     'initialized modules')
             elif not module_initialized and state_initialized:
-                # Here we need to avoid a tensor size missmatch
+                # Here we need to avoid a tensor size mismatch
                 # this is a regular tensor without a materialize
                 # method, so we can just resize for the load logic to copy
                 # the contents later to the correct device the module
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/nn/modules/lazy.py
```diff
@@ -90,24 +90,36 @@ def _lazy_load_hook(
...
+                    'Can\'t load initialized buffers in non-initialized '
+                    'modules and vice versa')
+            else:
+                # For compatibility with older snapshots
+                if key not in state_dict and not module_initialized:
+                    buffer = torch.Tensor([])
+                    state_dict[key] = buffer
 
         for name in self.lazy_parameter_names:
             # The parameter may not exist in the loaded ``state_dict`` if the
```
kmaehashi, line 108:
```suggestion
            # The parameter does not exist in the loaded ``state_dict`` if the
```

pytorch_pfn_extras/nn/modules/lazy.py
```diff
@@ -90,24 +90,36 @@ def _lazy_load_hook(
...
             # parameters (see comments of ``state_dict``).
             key = prefix + name
-            if key in state_dict:
-                # The model was serialized after initialization.
-                self.register_parameter(
-                    name, torch.nn.Parameter(state_dict[key]))
+            module_initialized = not isinstance(
+                getattr(self, name), UninitializedParameter)
+            state_initialized = key in state_dict
+            if module_initialized != state_initialized:
```
kmaehashi, line 115: I think the only case we have to take care of is this case:
```suggestion
            if module_initialized and not state_initialized:
```
as registering uninitialized parameters to optimizers will raise a warning.

pytorch_pfn_extras/nn/modules/lazy.py
```diff
@@ -90,24 +90,36 @@ def _lazy_load_hook(
...
+                raise RuntimeError(
+                    'Can\'t load initialized parameters in non-initialized '
+                    'modules and vice versa')
```
kmaehashi, line 118:
```suggestion
                    'Can\'t load uninitialized parameters in already-initialized '
                    'modules')
```

pytorch_pfn_extras/nn/modules/lazy.py
```diff
@@ -90,24 +90,36 @@ def _lazy_load_hook(
...
             else:
-                # The model was serialized before initialization.
-                param = UninitializedParameter()
-                self.register_parameter(name, param)
-                state_dict[key] = param
+                if key not in state_dict and not module_initialized:
```
kmaehashi, line 120:
```py
              elif not module_initialized and not state_initialized:
```

pytorch_pfn_extras/nn/modules/lazy.py
```diff
@@ -90,24 +90,36 @@ def _lazy_load_hook(
         for the details of the hook specification.
         """
         for name in self.lazy_buffer_names:
-            # Avoid shape mismatch error when loading an initialized buffer
-            # onto an uninitialized module instance.
-            self.register_buffer(name, state_dict[prefix + name])
+            key = prefix + name
```
kmaehashi, line 93: I don't think it's necessary to take care of buffers.

kmaehashi (changes requested):

kmaehashi:
Could you fix the comment in manager.py as well?

```
        The purpose of this hook is to adjust the current state and/or
        ``state_dict`` being loaded so that a module instance serialized in
        both un/initialized state can be deserialized onto both un/initialized
        module instance.
```

pytorch_pfn_extras/nn/modules/lazy.py
emcastillo: I agree as the modules with uninitialized buffers also have uninitialized parameters.

emcastillo:
Thank you!
PTAL

pytorch_pfn_extras/nn/modules/lazy.py
```diff
@@ -81,32 +81,27 @@ def _lazy_load_hook(
...
-            else:
-                # The model was serialized before initialization.
+            module_initialized = not isinstance(
+                getattr(self, name), UninitializedParameter)
+            state_initialized = key in state_dict
+            if module_initialized and not state_initialized:
+                raise RuntimeError(
+                    'Can\'t load uninitialized parameters in already '
+                    'initialized modules')
+            elif key not in state_dict and not module_initialized:
```
kmaehashi, line 103:
```suggestion
            elif not module_initialized and not state_initialized:
```

kmaehashi:
Thanks!
> Fixes #167
Is this correct? I think this PR does not resolve this issue.

emcastillo:
> Is this correct? I think this PR does not resolve this issue.

Yes, this PR ensures that the parameters are not recreated, so autoload loads the state in the correct parameters and devices.

pytorch_pfn_extras/nn/modules/lazy.py
```diff
@@ -81,32 +81,27 @@ def _lazy_load_hook(
...
+        are properly initialized.
 
         See comment in ``torch.nn.Module._register_load_state_dict_pre_hook``
         for the details of the hook specification.
         """
-        for name in self.lazy_buffer_names:
-            # Avoid shape mismatch error when loading an initialized buffer
-            # onto an uninitialized module instance.
-            self.register_buffer(name, state_dict[prefix + name])
-
```
kmaehashi, line 96: This needs to be reverted.

emcastillo: if we do `register_buffer` we will overwrite the device, so we won't solve the bug, we need to work a way around it

tests/pytorch_pfn_extras_tests/nn_tests/modules_tests/test_lazy.py
```diff
@@ -136,17 +136,42 @@ def test_save_load(self, init_src, init_dst):
...
 
         with tempfile.NamedTemporaryFile(delete=False) as f:
             torch.save(model_src.state_dict(), f.name)
-            model_dst.load_state_dict(torch.load(f.name))
+            if init_src != init_dst:
+                with pytest.raises(RuntimeError):
+                    model_dst.load_state_dict(torch.load(f.name))
+            else:
+                model_dst.load_state_dict(torch.load(f.name))
+            print(torch.load(f.name))
```
kmaehashi, line 162:
```suggestion
```

tests/pytorch_pfn_extras_tests/nn_tests/modules_tests/test_lazy.py
```diff
@@ -136,17 +136,42 @@ def test_save_load(self, init_src, init_dst):
...
+
+        # Ensure that if the model was initialized, the parameters are the same
+        # after loading a state dict
```
kmaehashi, line 165:
```suggestion
        # Ensure that if the model was initialized, the parameters and buffers
        # are the same after loading a state dict
```

pytorch_pfn_extras/nn/modules/lazy.py
```diff
@@ -88,6 +88,26 @@ def _lazy_load_hook(
         See comment in ``torch.nn.Module._register_load_state_dict_pre_hook``
         for the details of the hook specification.
         """
+        for name in self.lazy_buffer_names:
+            key = prefix + name
+            module_initialized = getattr(self, name).shape != (0,)
+            state_initialized = (
+                 key in state_dict and (state_dict[key].shape != (0,)))
```
kmaehashi, line 95:
```suggestion
            state_initialized = state_dict[key].shape != (0,)
```

tests/pytorch_pfn_extras_tests/nn_tests/modules_tests/test_lazy.py
emcastillo: just removed it :)

pytorch_pfn_extras/nn/modules/lazy.py
```diff
@@ -88,6 +88,26 @@ def _lazy_load_hook(
...
+            if module_initialized and not state_initialized:
+                raise RuntimeError(
+                    'Can\'t load non-initialized buffers in already '
+                    'initialized modules')
+            elif not module_initialized and state_initialized:
+                # Here we need to avoid a tensor size missmatch
```
kmaehashi, line 101:
```suggestion
                # Here we need to avoid a tensor size mismatch
```

pytorch_pfn_extras/nn/modules/lazy.py
```diff
@@ -88,6 +88,26 @@ def _lazy_load_hook(
...
+                # this is a regular tensor without a materialize
+                # method, so we can just resize for the load logic to copy
+                # the contents later to the correct device the module
+                # was moved to
+                getattr(self, name).resize_(state_dict[key].size())
+            elif key not in state_dict and not module_initialized:
```
kmaehashi, line 107: I think the buffer keys are always in the state dict.

kmaehashi (approved):
Thanks!! LGTM 👍🏼

kmaehashi:
LGTM! Test failures are PIL bug and not related to this PR.

Issue #193, opened 2021-06-18
CI failures in torch19


asi1024:
- [x] test_tensorboard_writing (linux, win): #194
- [x] test_empty_shared_dataset (win): #194
- [x] test_transform[tuple-None-indices53-None-True] (linux): #194 
- [x] test_annotate (linux, win) #196
- [x] test_apply_annotation (linux, win) #196
- [x] test_scoped_anchor (linux, win) #197 
- [x] test_export_testcase (linux, win) #197
- [x] test_tensorboard_writer (linux, win): #194
- [x] test_queue_writer (linux): #194

linux: https://ci.preferred.jp/pytorch-pfn-extras.torch19-linux/73786/
win: https://ci.preferred.jp/pytorch-pfn-extras.torch19-win/73789/

asi1024:
The UserWarning raise while importing `torch.utils.tensorboard` can be resolved by updating `setuptools` to the newest version.

asi1024:
Now all CIs are green!

PR #196, opened 2021-06-21
Fix name of nodes ckecked in test_annotate

asi1024:
Part of #193

### DIFF
--- a/tests/pytorch_pfn_extras_tests/onnx/test_annotate.py
+++ b/tests/pytorch_pfn_extras_tests/onnx/test_annotate.py
@@ -47,7 +47,9 @@ def test_annotate():
     assert 'Conv_1' in named_nodes
 
     assert 'MatMul_3' in named_nodes
+    assert 'Add_4' in named_nodes
     assert 'MatMul_6' in named_nodes
+    assert 'Add_7' in named_nodes
 
     node_conv_0_attrs = [a.name for a in named_nodes['Conv_0'].attribute]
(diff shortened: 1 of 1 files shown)


kmaehashi:
/test

Let's see if these tests pass with PyTorch 1.8 or 1.7. If fails I think it's ok to skip and just support the latest PyTorch in ppe.onnx.

@take-cheeze It would be great if you could take a look on diff whether this fix is reasonable :pray:

tests/pytorch_pfn_extras_tests/onnx/test_annotate.py
```diff
@@ -141,6 +155,7 @@ def _fn2(x):
     assert 'yyy' in node_attrs
 
 
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
```
asi1024, line 158: This DeprecationWarning seems to be raised only on the first call of `export_testcase`.

take-cheeze:
memo: CI failure of warning would be fixed by https://github.com/pfnet/pytorch-pfn-extras/pull/197

asi1024:
@take-cheeze @kmaehashi 
Fixed to use `Linear(..., bias=False)` to simplify computation graph.

take-cheeze (approved):
LGTM 👍

kmaehashi:
LGTM, thanks @asi1024 for fixing and @take-cheeze for reviewing!

PR #197, opened 2021-06-21
Add filterwarnings in some onnx tests

asi1024:
Part #193

### DIFF
--- a/tests/pytorch_pfn_extras_tests/onnx/test_annotate.py
+++ b/tests/pytorch_pfn_extras_tests/onnx/test_annotate.py
@@ -153,6 +153,7 @@ def test_apply_annotation():
 
 @pytest.mark.filterwarnings("ignore::DeprecationWarning")
 @pytest.mark.filterwarnings("ignore::torch.jit.TracerWarning")
+@pytest.mark.filterwarnings("ignore::UserWarning")
 def test_scoped_anchor():
     class Net(nn.Module):
         def __init__(self, anchor_mode='on'):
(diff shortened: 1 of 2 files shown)


kmaehashi:
Can we narrow down the filter a bit more? (maybe limit to torch.floor_divide)

tests/pytorch_pfn_extras_tests/onnx/test_annotate.py
```diff
@@ -142,6 +142,7 @@ def _fn2(x):
 
 
 @pytest.mark.filterwarnings("ignore::torch.jit.TracerWarning")
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
asi1024, line 145: Deprecation warning of `torch.floor_divide` raised from here: https://github.com/pytorch/pytorch/blob/v1.9.0/torch/nn/functional.py#L2358

tests/pytorch_pfn_extras_tests/onnx/test_export_testcase.py
```diff
@@ -62,6 +62,7 @@ def _helper(model, args, d, **kwargs):
     return output_dir
 
 
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
asi1024, line 65: An experimental warning for named tensors is raised, but this test looks not to use named tensor :thinking:
```
UserWarning: Named tensors and all their associated APIs are an experimental feature and subject to change. Please do not use them for anything important until they are released as stable. (Triggered internally at  /pytorch/c10/core/TensorImpl.h:1156.)
```

asi1024:
@kmaehashi Could you take another look?

kmaehashi:
Could you resolve conflicts? LGTM otherwise!

kmaehashi (approved):
Thanks!

PR #179, opened 2021-06-02
Make ExtensionsManager typed

asi1024:
Depends on ~#182, #183, #200 and #205~

PR #200, opened 2021-06-29
Fix `make_extension` to return `_CustomExtension` object

asi1024:
Fix https://github.com/pfnet/pytorch-pfn-extras/pull/179#issuecomment-870174101.


pytorch_pfn_extras/training/extension.py
```diff
@@ -126,8 +126,44 @@ def load_state_dict(self, to_load):
         pass
 
 
-def make_extension(trigger=None, default_name=None, priority=None,
-                   finalizer=None, initializer=None, on_error=None):
+class _CustomExtension(Extension):
```
kmaehashi, line 129: Better name it `_CallableExtension` to align with `_CallableTrigger` proposed in https://github.com/pfnet/pytorch-pfn-extras/pull/179 ?

pytorch_pfn_extras/training/extension.py
```diff
@@ -126,8 +126,44 @@ def load_state_dict(self, to_load):
...
+        return self._on_error(manager, exc, tb)
+
+
+def make_extension(
+        trigger=(1, 'iteration'),
+        default_name=None,
+        priority=PRIORITY_READER,
+        finalizer=lambda: None,
+        initializer=lambda manager: None,
+        on_error=lambda manager, exc, tb: None,
```
kmaehashi, line 165: Are these values defined? I think the default behavior should be implemented in one place (which is, `class Extension`)

kmaehashi (changes requested):
I think fix is needed to allow accepting callable in `manager.extend()`

kmaehashi (approved):
LGTM! :+1:

pytorch_pfn_extras/training/extension.py
```diff
@@ -126,8 +126,44 @@ def load_state_dict(self, to_load):
...
+        trigger=Extension.trigger,
+        default_name=None,
+        priority=Extension.priority,
+        finalizer=lambda: None,
+        initializer=lambda manager: None,
+        on_error=lambda manager, exc, tb: None,
+):
     """Decorator to make given functions into extensions.
 
     This decorator just adds some attributes to a given function. The value of
```
kmaehashi, line 169: Could you update docs?

pytorch_pfn_extras/training/manager.py
```diff
@@ -187,9 +187,9 @@ def start_extensions(self):
 
         # invoke initializer of each extension
         for _, entry in self.extensions:
-            initializer = getattr(entry.extension, 'initialize', None)
+            initializer = entry.extension.initialize
```
kmaehashi, line 190: Is this safe? I guess `manager.extend` currently accepts a callable

PR #207, opened 2021-07-15
[ONNX] Support stripped ONNX

disktnk:
When running `strip_large_tensor` with ONNX which has already stripped, occur an error

```
FileNotFoundError: [Errno 2] No such file or directory: '/path/{"type": "stripped", "average": 0.0002447886217851192, "variance": 0.004106185864657164}'
```

Because stripped meta information is written as "external data location" and `onnx.load` try to loading the external location path https://github.com/onnx/onnx/blob/685898bc6f3fd6eb16ba129ea990f96330537e38/onnx/external_data_helper.py#L47

This PR fixed the error, by skipped stripped tensor when (re)stripping.


kmaehashi:
I'm reviewing but would be great if you could double-check @take-cheeze 😄

pytorch_pfn_extras/onnx/strip_large_tensor.py
```diff
@@ -16,6 +16,22 @@ def is_large_tensor(tensor, threshold):
     return size > threshold
 
 
+def _is_stripped(tensor):
+    if len(tensor.external_data) == 0:
+        return False
+    external_value = None
+    for external_data in tensor.external_data:
+        if external_data.key == 'location':
+            external_value = external_data.value
```
kmaehashi, line 25: `location` won't appear twice? (correct me if I'm wrong)
```suggestion
            external_value = external_data.value
            break
```

pytorch_pfn_extras/onnx/strip_large_tensor.py
```diff
@@ -16,6 +16,22 @@ def is_large_tensor(tensor, threshold):
...
+    if not external_value:
+        return False
+    try:
+        external_value_dict = json.loads(external_value)
+        return external_value_dict.get('type', '') == 'stripped'
+    except Exception:
+        return False
```
kmaehashi, line 32: Let's use the narrower exception.

```suggestion
    except ValueError:
        # Invalid JSON, indicating `external_value` contains a file path
        return False
```

kmaehashi (changes requested):
LGTM except minor comments!

kmaehashi:
For later reference: https://github.com/onnx/onnx/blob/master/docs/ExternalData.md

take-cheeze (approved):

pytorch_pfn_extras/onnx/strip_large_tensor.py
disktnk: fixed

pytorch_pfn_extras/onnx/strip_large_tensor.py
disktnk: Before code could not consider multiple external data, so fixed to handle them.

take-cheeze:
Sorry I've miss-approved https://github.com/pfnet/pytorch-pfn-extras/pull/208

pytorch_pfn_extras/onnx/strip_large_tensor.py
```diff
@@ -16,6 +16,22 @@ def is_large_tensor(tensor, threshold):
     return size > threshold
 
 
+def _is_stripped(tensor):
+    if len(tensor.external_data) == 0:
+        return False
```
kmaehashi, line 21: I think this condition can now be removed.
```suggestion
```

kmaehashi (commented):
LGTM except for minor comment.

pytorch_pfn_extras/onnx/strip_large_tensor.py
disktnk: fixed

take-cheeze (approved):
LGTM 👍

kmaehashi (approved):

PR #208, opened 2021-07-15
[ONNX] Support annotated models but run without export

disktnk:
The model has `annotate` (or `anchor`) but run the model no under `ppe.onnx.export`, occurs an error.

```
>       for name, child_module in _annotation_init.model.named_children():
E       AttributeError: '_AnnotationInit' object has no attribute 'model'
```

This PR fix the error. `annotate` works only when `onnx.export` mode.

### DIFF
--- a/pytorch_pfn_extras/onnx/annotate.py
+++ b/pytorch_pfn_extras/onnx/annotate.py
@@ -1,8 +1,10 @@
+import contextlib
 import functools
 
 import onnx.helper
 import torch
 import torch.nn as nn
+import torch.onnx
 import torch.onnx.symbolic_registry as sym_reg
 
 
@@ -191,7 +193,9 @@ def annotate(**attrs):
     Args:
         attrs (dict): annotation parameters
     """
-    return _Annotation(**attrs)
+    if torch.onnx.is_in_onnx_export():
+        return _Annotation(**attrs)
+    return contextlib.nullcontext()
 
 
 def apply_annotation(fn, *args, **attrs):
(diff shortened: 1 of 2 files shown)


kmaehashi (approved):
LGTM!

pytorch_pfn_extras/onnx/annotate.py
```diff
@@ -191,7 +193,9 @@ def annotate(**attrs):
     Args:
         attrs (dict): annotation parameters
     """
-    return _Annotation(**attrs)
+    if torch.onnx.is_in_onnx_export():
```
take-cheeze, line 196: So `annotate`  and `scoped_anchor` would be ignored when it's not exporting onnx
👍

take-cheeze (approved):

take-cheeze:
CI failure seems to be caused by mypy. Should we wait https://github.com/pfnet/pytorch-pfn-extras/pull/209 ?

kmaehashi:
Sorry that log is difficult to understand. Currently mypy is enabled but errors are ignored. The true error in this PR is the flake8 issue:

```
/src/tests/pytorch_pfn_extras_tests/onnx/test_annotate.py:26:1: E303 too many blank lines (3)
```

tests/pytorch_pfn_extras_tests/onnx/test_annotate.py
```diff
@@ -22,6 +22,30 @@
 torch_version = version.Version(torch.__version__)
 
 
+
```
kmaehashi, line 25:
```suggestion
```

kmaehashi:
Thanks @disktnk and @take-cheeze!

PR #211, opened 2021-07-15
Increase memory for FlexCI

kmaehashi:


kmaehashi:
Merging as the test passed. I think this should prevent flaky failures regarding distributed tests.

PR #213, opened 2021-07-16
PlotReport: close plot when data is not set

corochann:
[close plot when data is not set]
when `a.has_data()` is false at PlotReport, it does not call `savefun` and it does not close figure instance `f`.
it causes to show warning, so I added code to close it even when a.has_data is False.

[thread safe close for matpltlib figure]

I think `plt.close()` may close other `fig` instance, if matplotlib is called in other thread/process. Explicitly close `fig` instance may be safer.

Also, I added to call `clf` method, to explicitly clear for better memory usage.
 - https://qiita.com/Masahiro_T/items/bdd0482a8efd84cdd270


emcastillo (approved):

emcastillo:
Thank you!!

PR #215, opened 2021-07-20
Fix cov command

emcastillo:
it was referring only to the tests dir


emcastillo:
seems like coverage report is not being copied :(

PR #223, opened 2021-07-31
Fix dead links on README.md 

yasuyuky:
I'm trying to test this library and looked at the README.md, but it seems that all the links in the documentation section of README.md are dead links.
Looking at the history, it looks like a simple omission, so I fixed it.


emcastillo (approved):
Thank you!!!

PR #232, opened 2021-08-02
Fix import error on Python 3.6 and 3.7

asi1024:


kmaehashi:
LGTM! Confirmed import failure has been resolved. Remaining failures will be fixed in separate PRs.

PR #233, opened 2021-08-02
Avoid using nullcontext to support Python 3.6

kmaehashi:


asi1024 (approved):

kmaehashi:
Confirmed all nullcontext related test failures disappear.

PR #234, opened 2021-08-02
Fix tests to avoid using args/kwargs attribute of Mock.call_args

kmaehashi:
This is needed as Mock.call_args.{args,kwargs} was added in Python 3.8.


kmaehashi:
Merging as test passed.

PR #240, opened 2021-08-04
Fix #238 by launching JoinableQueue when it actually becomes necessary

### LINKED ISSUE #238: The new ppe profiler introduces an unexpected ill-effect where users cannot set multiprocessing process start mode after importing ppe
I found out that importing the latest ppe will cause 

```python
>>> import multiprocessing as mp
>>> import pytorch_pfn_extras as ppe
>>> print(ppe.__version__)
>>> mp.set_start_method('forkserver')   # 'fork' or whatever

  File "<string>", line 1, in <module>
  File "/home/xxxx/.local/lib/python3.8/site-packages/pytorch_pfn_extras/__init__.py", line 10, in <module>
    from pytorch_pfn_extras import profiler  # NOQA
  File "/home/xxxx/.local/lib/python3.8/site-packages/pytorch_pfn_extras/profiler/__init__.py", line 1, in <module>
    from pytorch_pfn_extras.profiler._record import record  # NOQA
  File "/home/xxxx/.local/lib/python3.8/site-packages/pytorch_pfn_extras/profiler/_record.py", line 4, in <module>
    from pytorch_pfn_extras.profiler._time_summary import time_summary
  File "/home/xxxx/.local/lib/python3.8/site-packages/pytorch_pfn_extras/profiler/_time_summary.py", line 157, in <module>
    mp.set_start_method('forkserver')
  File "/usr/local/lib/python3.8/multiprocessing/context.py", line 243, in set_start_method
    raise RuntimeError('context has already been set')
RuntimeError: context has already been set
```

What is actually happening inside is very similar to this. 
https://github.com/numba/numba/pull/5724
Only the difference between ppe is that ppe creates `multiprocessing.JoinableQueue` instead of `get_context()` as in numba. In either case the process start method is implicitly fixed to "fork" right then and we can no longer change it anymore during the process lifetime, no matter which mode actually we want.

This happens at **import time** of ppe, specifically in instantiation of `TimeSummary`, not when the first time actually use it.
https://github.com/pfnet/pytorch-pfn-extras/blob/94a0d5092731ca94cc2deefaf5a55d037ec30dcc/pytorch_pfn_extras/profiler/_time_summary.py#L155
https://github.com/pfnet/pytorch-pfn-extras/blob/94a0d5092731ca94cc2deefaf5a55d037ec30dcc/pytorch_pfn_extras/profiler/_time_summary.py#L18

### Simple workaround

As written in https://github.com/numba/numba/pull/5724, if we set start method before importing ppe, it works fine.

```python
>>> import multiprocessing as mp
>>> mp.set_start_method('forkserver')
>>> import pytorch_pfn_extras as ppe
>>> print(ppe.__version__)
```

### Fundamental fix

I think we need to delay instantiation of _CPUWorker (or at least creation of `JoinableQueue`) to when it actually becomes necessary.

belltailjp:
Attempt for fixing #238


emcastillo (approved):
LGTM!!!!!

belltailjp:
Thanks to @HiroakiMikami 's contribution, I'd like to delegate to review phase! @emcastillo 

I wanted to write a test as in [the reverted commit](https://github.com/pfnet/pytorch-pfn-extras/pull/240/commits/10ccf211bca968eb0d4e08fe51c63fff152e5c01), but this test case heavily relies on the execution order, so I think it's quite difficult unless there's a good idea to make sure to run this particular test case early...

emcastillo:
Testing this per se its difficult, we will ad a new folder called `integration_tests` that will be run in its own pytest process.

emcastillo (approved):

emcastillo:
seems that windows is causing some datarace with the initialization ..

belltailjp:
I seem to have reproduced the CI failure on my local environment (although I'm on Linux).

Apparently if the report it sent immediately after launching CPUWorker, first some reports are missed.

```python
# repro.py
import time

with ppe.profiler.time_summary.report("iter-time"):
    time.sleep(0.1)
with ppe.profiler.time_summary.report("iter-time"):
    time.sleep(0.2)

with ppe.profiler.time_summary.summary() as s:
    print(s.compute_mean())
```

```bash
% python repro.py
{}
# where I expect sth like {'iter-time': 0.15000}
```

I am not actually sure why this happens, because JoinableQueue itself can handle items immediately added.
```python
import multiprocessing as mp
q = mp.JoinableQueue(1000)
q.put('hello')

v = q.get()
print(v)

# This will print "hello", which is expected.
```

I guess the first few samples are missed because the queue or something is not ready right after they are created, so I tried to add a sufficient sleep, which works as expected.
```python
+ time.sleep(1.0)
  with ppe.profiler.time_summary.summary() as s:
      ...
```
```bash
% python repro.py
{'iter-time': 0.15025246143341064}
```

Adding `time_summary.wait()` looks also fine.
```python
+ ppe.profiler.time_summary.wait()
  with ppe.profiler.time_summary.summary() as s:
      ...
```
So maybe I can add `self.wait()` at [`TimeSummary.report`](https://github.com/pfnet/pytorch-pfn-extras/blob/94a0d5092731ca94cc2deefaf5a55d037ec30dcc/pytorch_pfn_extras/profiler/_time_summary.py#L147), but
- I don't exactly understand why this can fix
- May adding `wait()` every time we call `report()` cause of extra overhead at the main process?

kmaehashi:
Closing as we merged #260. Thanks again! :pray:

PR #244, opened 2021-08-05
Typing: reporting.py

asi1024:


pytorch_pfn_extras/reporting.py
```diff
@@ -72,20 +73,25 @@ class Reporter:
...
-    def __enter__(self):
+    def __enter__(self) -> None:
         """Makes this reporter object current."""
         _get_reporters().append(self)
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(
+            self,
+            exc_type: Optional[Type[Exception]],
+            exc_value: Optional[Exception],
```
kmaehashi, line 87: Should it be `BaseException`?

kmaehashi (changes requested):

pytorch_pfn_extras/reporting.py
asi1024: Fixed. PTAL!

PR #247, opened 2021-08-06
Add DistributedSubsetSampler

emcastillo:
Originally submitted by @HiroakiMikami


kmaehashi:
As discussed internally let's avoid using sampler and add function to compute indices for each GPU.

PR #248, opened 2021-08-06
Update setup.py

shigengtian:
When I use pytroch-pfn-extras

AttributeError: module 'pytorch_pfn_extras.training.extensions' has no attribute 'PrintReportNotebook'  occured

```python
try:

    from pytorch_pfn_extras.training.extensions.print_report_notebook import PrintReportNotebook  # NOQA
    from pytorch_pfn_extras.training.extensions.progress_bar_notebook import ProgressBarNotebook  # NOQA

    _ipython_module_available = True
except ImportError:
    _ipython_module_available = False
```

I found that PrintReportNotebook, and ProgressBarNotebook need requirements. So I add them to setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ setuptools.setup(
                 'development in PyTorch.',
     author='Preferred Networks, Inc.',
     license='MIT License',
-    install_requires=['numpy', 'torch'],
+    install_requires=['numpy', 'torch', 'ipython', 'ipywidgets'],
     extras_require={
         'test': ['pytest'],
         'onnx': ['onnx'],


emcastillo:
Thank you for your PR,
Actually we would like to keep the requirements as they are, 
How about changing the `ImportError` to `AttributeError`?
Thank you!

shigengtian:
Maybe show some messages to the user will be kindly.  When I use pytorch-pfn-extras in a new env. 

If I use PrintReportNotebook, This error will occure
```
  File "/media/workstation/data_disk/shi_backup/seti/base.py", line 364, in <module>
    "PrintReportNotebook": ppe_exts.PrintReportNotebook,
AttributeError: module 'pytorch_pfn_extras.training.extensions' has no attribute 'PrintReportNotebook'
```

I must check the code.

print_report_notebook.py
``` python
import sys

from IPython.core.display import display
from ipywidgets import HTML

from pytorch_pfn_extras.training.extensions.print_report import PrintReport


class PrintReportNotebook(PrintReport):

....

```
and  run 
```shell
pip install  IPython and ipywidgets.
```

every time.

emcastillo:
Are you calling these in jupyter notebook?
they can only be used in jupyter

emcastillo:
we discussed this internally and we have decided to make these extensions no-op when running outside ipython. But show a warning to the user.
Does this satisfy your use case?

Sorry but we dont want to force users to install optional dependences😖

kmaehashi:
Thank you for raising this discussion, @shigengtian!

It was our fault that we did not document these requirements. As shown in the installation guide, there are several optional dependencies to run specific PPE features: https://github.com/pfnet/pytorch-pfn-extras#installation
This is because we want to keep PPE dependency minimal.

I understand that you want to run the same code both inside and outside of Jupyter. (is this correct?)
In that case as commented by Emilio, making these extensions no-op when running outside of Jupyter sounds the right way to go. Of course, we also have to update our README to explicitly document these package dependencies.

shigengtian:
Thank you very much. I have understood the reason.

kmaehashi:
I've raised separate issues to handle with these issues. #253, #254

PR #252, opened 2021-08-06
Typing: logging

asi1024:


pytorch_pfn_extras/logging.py
```diff
@@ -31,12 +37,12 @@ def _configure_logging(*, filename=None, level='ERROR', format=_logger_format):
...
 
 
-def _get_root_logger():
+def _get_root_logger() -> Optional[logging.Logger]:
     """Returns a logger to be used by pytorch-pfn-extras."""
     return _logger
 
 
-def get_logger(name):
+def get_logger(name: str) -> Optional[logging.Logger]:
```
emcastillo, line 45: I think this is not `Optional`, it always returns a logger, same for the above one, they are initialized in `__init__`

asi1024: Fixed.

PR #255, opened 2021-08-09
Export record_function and record_iterable to public

HiroakiMikami:
Add `record_function` and `record_iterable` to public. These two functions became private in changes of #226.


kmaehashi:
Thanks!
/test

PR #260, opened 2021-08-10
Defer initialization of profiler

### LINKED ISSUE #238: The new ppe profiler introduces an unexpected ill-effect where users cannot set multiprocessing process start mode after importing ppe

kmaehashi:
Tried to fix #238.
This is based on the proposal of @belltailjp and @HiroakiMikami (#240) :pray:

This PR defers initialization (thread creation and multiprocessing context initialization) to make `import pytorch_pfn_extras` harmless.

### DIFF
--- a/pytorch_pfn_extras/profiler/_time_summary.py
+++ b/pytorch_pfn_extras/profiler/_time_summary.py
@@ -176,6 +176,7 @@ class TimeSummary:
         self._cpu_worker.finalize()
         if self._cuda_worker is not None:
             self._cuda_worker.finalize()
+        self._initialized = False
 
     def synchronize(self):
         self.initialize()
--- a/pytorch_pfn_extras/profiler/_time_summary.py
+++ b/pytorch_pfn_extras/profiler/_time_summary.py
@@ -1,9 +1,9 @@
-import atexit
 from contextlib import contextmanager
+import os
 import time
 from typing import Tuple
-from threading import Lock, Thread
-from queue import Queue
+import threading
+import queue
 import multiprocessing as mp
 import torch
 from pytorch_pfn_extras.reporting import DictSummary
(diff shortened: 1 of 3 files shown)


emcastillo:
fixed conflicts and flake8

emcastillo (approved):

PR #264, opened 2021-08-11
Book-keeping of future-like objects

emcastillo:
Before, async iterations returned `None`,
Now we allow to return arbitrary objects that are passed to the runtime check

### DIFF
--- a/pytorch_pfn_extras/handler.py
+++ b/pytorch_pfn_extras/handler.py
@@ -200,6 +200,7 @@ class Handler(BaseHandler):
         self._train_report_keys = config.pop('train_report_keys', [])
         self._async = config.pop('async', False)
         self.pending_iters = defaultdict(list)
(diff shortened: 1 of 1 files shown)


pytorch_pfn_extras/handler.py
```diff
@@ -314,13 +315,15 @@ def train_validation_end(self, trainer, evaluator):
     def _complete_train_step(self, trainer, outs, block, sn, sm, rt):
         idx, batch, cback = self.pending_iters[sn][0]
         self.pending_iters[sn] = self.pending_iters[sn][1:]
+        self.futures[sn] = self.futures[sn][1:]
```
emcastillo, line 318: Merge this list and objects in a single object with `_PendingAsyncIteration` class?

kmaehashi:
Could you resolve conflicts? :bow:

PR #266, opened 2021-08-12
Register atexit handler

kmaehashi:
Follow-up: #260


kmaehashi:
Let me test several times... 
/test

emcastillo (approved):

PR #270, opened 2021-08-13
Typing: Runtime

asi1024:


pytorch_pfn_extras/runtime/_runtime.py
```diff
@@ -1,7 +1,20 @@
+from typing import (
+    Any, Dict, Generator, List, Optional, Sequence, Tuple, Union, TYPE_CHECKING
+)
+
 import torch
 
 _RUNTIME_TAG_NAME = '_ppe_runtime'
 
+DeviceLike = Union[str, torch.device]
+Batch = Union[torch.Tensor, List[torch.Tensor],
```
kmaehashi, line 10: Could you change to Any as discussed?

pytorch_pfn_extras/runtime/_runtime.py
```diff
@@ -1,7 +1,20 @@
...
+              Tuple[torch.Tensor, ...], Dict[str, torch.Tensor]]
+Loader = Sequence[Batch]
```
kmaehashi, line 12: I think this should be `Iterable` as `torch.utils.data.DataLoader` is not a sequence.

kmaehashi (changes requested):

pytorch_pfn_extras/runtime/_runtime.py
emcastillo: Agree

emcastillo (approved):

PR #271, opened 2021-08-13
Typing: `_RuntimeRegistry` and `ppe.to`

asi1024:
Depends on #270

### DIFF
--- a/pytorch_pfn_extras/runtime/_registry.py
+++ b/pytorch_pfn_extras/runtime/_registry.py
@@ -1,4 +1,4 @@
-from typing import Dict, Type, TYPE_CHECKING
+from typing import Dict, Type
 
 import torch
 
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/runtime/_registry.py
```diff
@@ -1,15 +1,25 @@
+from typing import Dict, Type, TYPE_CHECKING
+
 import torch
 
+if TYPE_CHECKING:
```
kmaehashi, line 5: Question: Is this `if` condition needed?

pytorch_pfn_extras/runtime/_to.py
```diff
@@ -1,10 +1,22 @@
+from typing import Any, Dict, Optional, Type, TypeVar, TYPE_CHECKING
+
 import torch
 
 import pytorch_pfn_extras as ppe
 
+if TYPE_CHECKING:
```
kmaehashi, line 7: ditto

kmaehashi:
LGTM!
/test

PR #272, opened 2021-08-13
Do `autocast` only on forward

emcastillo:
Since torch 1.8 the autocast & checkpointing seems to be fixed, so we apply it only to the forward pass.

### DIFF
--- a/pytorch_pfn_extras/handler.py
+++ b/pytorch_pfn_extras/handler.py
@@ -27,6 +27,9 @@ def torch_autocast(enabled=True):
         yield
 
 
+_is_torch_17 = torch.__version__.split('.')[1] == '7'
+
+
 class BaseHandler:
 
     def __init__(self, logic, options, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/handler.py
```diff
@@ -27,6 +27,9 @@ def torch_autocast(enabled=True):
         yield
 
 
+_is_torch_17 = torch.__version__.split('.')[1] == '7'
```
emcastillo, line 30: This is horrible, minimum supported is 1.7 so I just check against it like this ...

kmaehashi:
As we now dropped support for PyTorch 1.7 (#307) let's remove the version check :smile:
Also please resolve conflicts :pray:

emcastillo:
PTAL!

kmaehashi:
Could you resolve conflicts and update docstrings? (`autocast` config moved to Logic)

emcastillo:
rebased! PTAL!

kmaehashi:
I've pushed docstring fixes!
/test

PR #273, opened 2021-08-13
Typing: metrics

asi1024:


pytorch_pfn_extras/training/metrics.py
```diff
@@ -11,18 +17,20 @@ class AccuracyMetric:
     .. seealso:
        :func:`pytorch_pfn_extras.engine.create_evaluator`
     """
-    def __init__(self, label_key, output_key):
+    def __init__(self, label_key: str, output_key: str) -> None:
```
toslunar, line 20: Now we can remove `(str)` from the docstring.

PR #280, opened 2021-08-20
Model comparer

emcastillo:
Same as outputs comparer but this does it for model parameters,
It refactors the compares to create a common class, and then specializes it for both outputs and models.

Models allow to specify parameters to be compared using regular expressions

TODO: docstring, increase tests


kmaehashi:
@emcastillo Could you rebase this PR?

asi1024 (approved):

PR #282, opened 2021-08-23
skip assert when a model returns tensor

yuyu2172:
The assertion triggers false positively.

This happens when I use a loss function that returns tensor (not tuple, dict, ...).

### DIFF
--- a/pytorch_pfn_extras/handler.py
+++ b/pytorch_pfn_extras/handler.py
@@ -584,7 +584,8 @@ class Logic(BaseLogic):
         to_back_outs = outs
         if self._grad_scaler is not None:
             to_back_outs = self._normalize_outputs(outs)
-            assert len(outs) == 1, "loss scaling with multiple outputs is not supported"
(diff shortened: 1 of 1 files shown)


pytorch_pfn_extras/handler.py
```diff
@@ -584,7 +584,8 @@ def train_step(self, models, optimizers, batch_idx, batch):
         to_back_outs = outs
         if self._grad_scaler is not None:
             to_back_outs = self._normalize_outputs(outs)
-            assert len(outs) == 1, "loss scaling with multiple outputs is not supported"
+            if not isinstance(outs, torch.Tensor):
+                assert len(outs) == 1, "loss scaling with multiple outputs is not supported"
```
emcastillo, line 588:
```suggestion
             assert len(to_back_outs) == 1, "loss scaling with multiple outputs is not supported"
```

I think that `_normalize_outputs` will create a dict with the tensor, so this should be enough

emcastillo (approved):

PR #287, opened 2021-08-25
Typing: Evaluator

asi1024:
Depends on https://github.com/pfnet/pytorch-pfn-extras/pull/274.

PR #269, opened 2021-08-12
Fix `Handler.eval_step` and make `Logic` and `Handler` typed

asi1024:

PR #288, opened 2021-08-25
Typing: Trainer

asi1024:
Depends on https://github.com/pfnet/pytorch-pfn-extras/pull/287 and #269.


asi1024:
I notice that this PR depends also on #269.

PR #292, opened 2021-08-29
Make manager state consistent and allow to synchronize in the handler

emcastillo:
we should load the state with `exeuction=iteration` since the on-going executions do not exists when restoring the state, and this leads to an inconsistent state.

cc @HiroakiMikami

### DIFF
--- a/pytorch_pfn_extras/handler.py
+++ b/pytorch_pfn_extras/handler.py
@@ -5,6 +5,7 @@ import torch
 
 import pytorch_pfn_extras as ppe
 from pytorch_pfn_extras import reporting
+from pytorch_pfn_extras.training import trigger as trigger_module
 
 
 _amp_enabled = False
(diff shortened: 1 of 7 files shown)


kmaehashi:
* Could you update tests as well?
* States of extensions with `is_async = True` flag also needs this.

emcastillo:
PTAL

pytorch_pfn_extras/handler.py
```diff
@@ -205,6 +206,9 @@ def __init__(self, logic, entry_runtime, options):
                     Default is an empty list.
                 * ``'async'`` (bool):
                     If ``True``, async mode is enabled. Default is ``False``.
+                * ``'sync_trigger'`` (tuple or Trigger):
+                    Trigger used to synchronize iterations in
+                    asynchronous mode. Default is ``None``.
```
kmaehashi, line 211: Do we need to pass the trigger manually? I was thinking that `_synchronize_train` is automatically called for all `needs_model_state = True` extensions.

emcastillo:
I did an alternative implementation where I add a `needs_sync` attr to the extensions and the target object to be synced (manager does not have direct access to trainer/handler), This needs to increase the public APIs of handler and trainer.

Personally I think that passing the trigger as a handler option is much cleaner 😅

PR #293, opened 2021-08-30
Add util to handle internal values as onnx outputs

HiroakiMikami:


pytorch_pfn_extras/onnx/as_output.py
```diff
@@ -0,0 +1,107 @@
...
+            for i, v in enumerate(node.output):
+                if v in old_name_to_new_name:
+                    node.output[i] = old_name_to_new_name[v]
+
+        for v in onnx_graph.graph.input:
+            if v.name in old_name_to_new_name:
+                v.name = old_name_to_new_name[v.name]
+
+
+_outputs: Optional[_Outputs] = None
```
emcastillo, line 56: Instead of having this global variable, would it be possible to turn `_start_trace` and `_stop_trace` into a context manager?

emcastillo (changes requested):

pytorch_pfn_extras/onnx/as_output.py
HiroakiMikami: We can use context manager instead of two separate functions (and I will do refactoring later). But we cannot remove the global variable because it's used in `as_output` function. `as_output` function will be invoked in the user code.

emcastillo: If we have the context manager saving the `_outputs` in thread local storage, I think it is easier for the `as_output` function to retrieve it without having a global variable.
What do you think?

HiroakiMikami: I see. I think you are right, and I'll fix the implementation.

tests/pytorch_pfn_extras_tests/onnx/test_as_output.py
```diff
@@ -0,0 +1,99 @@
...
+            h = self.linear(h)
+            return h
+
+    model = Net()
+    x = torch.ones((1, 1, 32, 32))
+    y = model(x)
+    assert y.shape == (1, 6, 30, 20)
+
+
+
```
emcastillo, line 42:
```suggestion
```

pytorch_pfn_extras/onnx/_as_output.py
```diff
@@ -0,0 +1,95 @@
...
+    def add_outputs_to_model(self, onnx_graph: onnx.ModelProto) -> None:
+        if len(self.values) == 0:
+            return
+
+        old_name_to_new_name = {}
+        n_output = len(onnx_graph.graph.output)
+        assert n_output >= len(self.values)
+
+        # Rename last len(self.values) outputs
+        for i, additional_output in enumerate(onnx_graph.graph.output[-len(self.values):]):
```
emcastillo, line 37:
```suggestion
        for i, additional_output in enumerate(
                onnx_graph.graph.output[-len(self.values):]):
```

emcastillo (approved):

PR #314, opened 2021-09-07
Update trainer docs

kmaehashi:
Closes #295 and #290

PR #317, opened 2021-09-07
Copy args inside Handler and Logic

### LINKED ISSUE #308: Trainer: copy config in Handler
Handler should not modify config dict given by user.

emcastillo:
Fixes #308 

In order to do the validations inside the classes too, this PR superseedes #291

### DIFF
--- a/pytorch_pfn_extras/engine.py
+++ b/pytorch_pfn_extras/engine.py
@@ -107,11 +107,13 @@ def create_trainer(
         writer: Optional['writing.Writer'] = None,
         evaluator: Optional['_Evaluator'] = None,
         device: 'DeviceLike' = 'cpu',
-        options: Optional[Dict[str, Any]] = None,
         logic: Optional[handler_module.Logic] = None,
         transform_model: Callable[
             [str, torch.nn.Module], torch.nn.Module] = default_transform_model,
         handler_class: Optional[Type[handler_module.BaseHandler]] = None,
+        handler_options: Optional[Dict[str, Any]] = None,
+        logic_options: Optional[Dict[str, Any]] = None,
+        runtime_options: Optional[Dict[str, Any]] = None,
 ) -> '_Trainer':
     """Creates a trainer object.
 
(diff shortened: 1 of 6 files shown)


kmaehashi:
* Could you update constructors and code of Runtime/Logic/Handler to use `option` instead of `config`?
* Please keep backward compatibility with the previous version like #291.

pytorch_pfn_extras/engine.py
```diff
@@ -138,33 +140,40 @@ def create_trainer(
...
+        handler_options (dict):
+            Options that are set to the handler object. See the documentation
+            of `ppe.handler.Handler` for details.
+        logic_options (dict):
+            Options that arte set to the logic object. When using the default
+            logic class, See the documentation of `ppe.handler.Logic`
+            for details.
+        handler_options (dict):
+            Options that are set to the runtime object. See the documentation
+            of `ppe.handler.Handler` for details.
```
kmaehashi, line 158:
```suggestion
        handler_options:
            Options that are set to the handler object. See the documentation
            of `ppe.handler.Handler` for details.
        logic_options:
            Options that are set to the logic object. When using the default
            logic class, See the documentation of `ppe.handler.Logic`
            for details.
        runtime_options:
            Options that are set to the runtime object. See the documentation
            of `ppe.handler.PyTorchRuntime` for details.
```

emcastillo:
> Please keep backward compatibility with the previous version like Separate options argument into logic_options and runtime_options #291.

This is hard because if we pass an unified dict of logic/handler options then checking if all the options were not consumed in either of the class is going to be difficult.

pytorch_pfn_extras/engine.py
emcastillo: All the above params in the doc string have their type specified, should we remove all?

kmaehashi: Updated suggestion, let's keep it as-is and discuss in #314 about that :smile:

emcastillo:
PTAL!

emcastillo:
rebased and removed types from docstring

PR #319, opened 2021-09-07
Mark some lazy modules as deprecated

kmaehashi:


emcastillo (approved):
LGTM! 
Should we raise a warning on linear and conv?

kmaehashi:
Test passed :tada:

PR #321, opened 2021-09-08
Backprop only when grad is enabled

emcastillo:
In some cases we want to skip the backprop call if the outputs didn't have grad_fn enabled.


kmaehashi (approved):
LGTM!

pytorch_pfn_extras/handler.py
```diff
@@ -630,6 +630,7 @@ def _backward(self, outputs):
             # This is to avoid errors when the trained models returns
             # tensors others than scalars
             if isinstance(v, torch.Tensor) and (
+                v.grad_fn is not None and
```
kmaehashi, line 633:
```suggestion
            if isinstance(v, torch.Tensor) and v.grad_fn is not None and (
```

Issue #320, opened 2021-09-07
Trainer: Split handler and logic into different file


kmaehashi:
`handler.py` is too long and needs refactoring.

PR #330, opened 2021-09-09
Move Logic definition above Handler

asi1024:


kmaehashi:
Why not fix #320 then? 😄

PR #340, opened 2021-09-13
Add `ppe.nn.Ensure` 

### LINKED ISSUE #147: Module to validate a shape and dtype?
Modules like `tf.ensure_shape` can be useful in PyTorch too.
https://www.tensorflow.org/api_docs/python/tf/ensure_shape

emcastillo:
Closes #147

### DIFF
--- a/pytorch_pfn_extras/nn/modules/ensure_shape.py
+++ b/pytorch_pfn_extras/nn/modules/ensure_shape.py
@@ -20,7 +20,7 @@ class Ensure(torch.nn.Module):
     def __init__(
             self,
             *,
-            shape: Optional[Tuple[int]] = None,
+            shape: Optional[Tuple[Optional[int]]] = None,
             dtype: Optional[torch.dtype] = None,
             broadcastable: bool = False,
             can_cast: bool = False,
@@ -93,7 +93,7 @@ class Ensure(torch.nn.Module):
 
 def ensure(
         tensor: torch.Tensor,
-        shape: Optional[Tuple[int]] = None,
+        shape: Optional[Tuple[Optional[int]]] = None,
         dtype: Optional[torch.dtype] = None,
         broadcastable: bool = False,
         can_cast: bool = False
--- a/pytorch_pfn_extras/nn/modules/ensure_shape.py
+++ b/pytorch_pfn_extras/nn/modules/ensure_shape.py
@@ -50,7 +50,7 @@ class EnsureShapeAndDtype(torch.nn.Module):
         l_1, l_2 = len(shape_1), len(shape_2)
         if l_1 != l_2:
             # we would like to do checks like the below one
-            # but torchscript converts everything to vectors
+            # but torchscript converts everything to tensors
             # shape_2 = (1,) * (l_1 - l_2) + shape_2
             if l_1 < l_2:
                 shape_1 = torch.cat([torch.ones(l_2 - l_1), shape_1])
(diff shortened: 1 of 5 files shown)


kmaehashi:
* How about supporting `None` in shape, like `ensure_shape`? I think users are likely to want to allow any size for the batch dimension.
* Maybe better having a wrapper API `ppe.ensure(tensor, *, shape=None, dtype=None, ...)`?
* Could you add tests to run these checks under jit?

emcastillo:
> How about supporting None in shape, like ensure_shape? I think users are likely to want to allow any size for the batch dimension.

I think it is currently supported, shape should be optional.

> Maybe better having a wrapper API ppe.ensure(tensor, *, shape=None, dtype=None, ...)?

This is nice to add, but the module should be kept too so we can use it inside `torch.nn.Sequential`.

> Could you add tests to run these checks under jit?

Sure, I wonder if it will be directly exported to onnx identity. Conditionals will be ignored in the exported module.
But they should be triggered while exporting.

kmaehashi:
> I think it is currently supported, shape should be optional.

Ah, I meant, e.g. `tf.ensure_shape(x, [None, 3])` to mark the first dimension unknown (allow any value)

emcastillo:
seems nice :)

emcastillo:
PTAL!

tests/pytorch_pfn_extras_tests/nn_tests/modules_tests/test_ensure_shape.py
```diff
@@ -0,0 +1,162 @@
...
+        ((2, 4), (4,)),
+        ((2, 3, 4), (1, 4)),
+    ])
+    def test_broadcastable_shape(self, shape_t, shape_c):
+        tensor = torch.zeros(shape_t)
+        module = EnsureShapeAndDtype(shape_c, broadcastable=True)
+        module(tensor)
+
+    @pytest.mark.parametrize('shape_t, shape_c', [
+        ((3,), (2,)),
```
kmaehashi, line 47: Better to test 0-dim case here?

kmaehashi (commented):
LGTM except for

emcastillo:
Addressed review comments! thanks!

pytorch_pfn_extras/nn/modules/ensure_shape.py
```diff
@@ -0,0 +1,115 @@
...
+           provided one.
+       broadcastable: Check if the shapes are compatible using broadcasting
+           rules.
+       can_cast: Check if the input tensor can be casted to the provided type.
+    """
+
+    def __init__(
+            self,
+            *,
+            shape: Optional[Tuple[int]] = None,
```
kmaehashi, line 23: mypy failed in CI as `shape` may contain None.

```suggestion
            shape: Optional[Tuple[Optional[int]]] = None,
```

pytorch_pfn_extras/nn/modules/ensure_shape.py
```diff
@@ -0,0 +1,115 @@
...
+                        f' {self._dtype}')
+            else:
+                raise ValueError(
+                    f'Expected {self._dtype}, input dtype is {input.dtype}')
+        return input
+
+
+def ensure(
+        tensor: torch.Tensor,
+        shape: Optional[Tuple[int]] = None,
```
kmaehashi, line 96:
```suggestion
        shape: Optional[Tuple[Optional[int]]] = None,
```

kmaehashi (changes requested):
LGTM except for minor comment.
Could you add new APIs to https://github.com/pfnet/pytorch-pfn-extras/blob/master/docs/source/reference/index.rst ?

kmaehashi:
I've added an example in 8119f8f. If it's ok for you, please proceed to merge!

kmaehashi (approved):

PR #343, opened 2021-09-15
Typing: manager.py

asi1024:


emcastillo:
```pytorch_pfn_extras/training/manager.py:668: error: Argument 1 to "len" has incompatible type "Union[DataLoader[Any], Iterable[Any], None]"; expected "Sized"  [arg-type]
```

pytorch_pfn_extras/training/manager.py
```diff
@@ -652,7 +665,7 @@ def set_reporter_on_iter(engine: Engine) -> None:
 
         @self.engine.on(Events.STARTED)
         def set_training_started(engine: Engine) -> None:
-            iters_per_epoch = len(engine.state.dataloader)
+            iters_per_epoch = len(engine.state.dataloader)  # type: ignore[arg-type]
```
asi1024, line 668: note: The actual type of `engine.state.dataloader` is `Union[DataLoader[Any], Iterable[Any], None]`, so the dataloader is not guaranteed to be `Sized`.

emcastillo (approved):
LGTM

PR #344, opened 2021-09-16
Remove `examples` from `pyproject.toml`

emcastillo:
It is quite annoying to have to go through all these known errors when looking the mypy output for other files


kmaehashi:
How about removing example from pyproject.yaml?

emcastillo:
Do we want to fix annotations of these files?

kmaehashi:
I'm thinking of checking examples and test codes with￼ allow-untyped-defs so that we can verify the correctness of types in PPE code base, without adding type annotations to examples and tests.

PR #364, opened 2021-09-28
Fix strip_large_tensor when tensor doesn't have raw_data

impactaky:
onnx.external_data_helper require raw_data field

### DIFF
--- a/pytorch_pfn_extras/onnx/strip_large_tensor.py
+++ b/pytorch_pfn_extras/onnx/strip_large_tensor.py
@@ -42,7 +42,7 @@ def _strip_raw_data(tensor):
         tensor.raw_data = onnx.numpy_helper.from_array(arr, tensor.name).raw_data
     onnx.external_data_helper.set_external_data(tensor,
                                                 location=json.dumps(meta_dict),
-                                                length=len(tensor.raw_data))
+                                                length=arr.nbytes)
     tensor.data_location = onnx.TensorProto.EXTERNAL
     tensor.ClearField('raw_data')
     tensor.ClearField('float_data')
--- a/pytorch_pfn_extras/onnx/strip_large_tensor.py
+++ b/pytorch_pfn_extras/onnx/strip_large_tensor.py
@@ -38,6 +38,8 @@ def _strip_raw_data(tensor):
     meta_dict['type'] = "stripped"
     meta_dict['average'] = float(numpy.average(arr))
     meta_dict['variance'] = float(numpy.var(arr))
+    if not tensor.HasField("raw_data"):
+        tensor.raw_data = onnx.numpy_helper.from_array(arr, tensor.name).raw_data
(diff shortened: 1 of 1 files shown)


kmaehashi:
/test


LGTM!
Also asking @take-cheeze for review 🙏

pytorch_pfn_extras/onnx/strip_large_tensor.py
```diff
@@ -38,6 +38,8 @@ def _strip_raw_data(tensor):
     meta_dict['type'] = "stripped"
     meta_dict['average'] = float(numpy.average(arr))
     meta_dict['variance'] = float(numpy.var(arr))
+    if not tensor.HasField("raw_data"):
+        tensor.raw_data = onnx.numpy_helper.from_array(arr, tensor.name).raw_data
     onnx.external_data_helper.set_external_data(tensor,
                                                 location=json.dumps(meta_dict),
                                                 length=len(tensor.raw_data))
```
take-cheeze, line 45: Instead of reconstructing `raw_data` how about reusing the `np.ndarray` field? :
```suggestion
                                                length=arr.nbytes)
```

pytorch_pfn_extras/onnx/strip_large_tensor.py
```diff
@@ -38,9 +38,11 @@ def _strip_raw_data(tensor):
     meta_dict['type'] = "stripped"
     meta_dict['average'] = float(numpy.average(arr))
     meta_dict['variance'] = float(numpy.var(arr))
+    if not tensor.HasField("raw_data"):
+        tensor.raw_data = onnx.numpy_helper.from_array(arr, tensor.name).raw_data
```
take-cheeze, line 42: Now this code would be unnecessary since we don't use `tensor.raw_data` anymore:
```suggestion
```

take-cheeze (approved):
Thank you for the fix! 👍

pytorch_pfn_extras/onnx/strip_large_tensor.py
kmaehashi: @impactaky Could you check this comment?
We are planning to cut v0.5.1 tomorrow, so if you would like to have this PR in the release, it would be great if you could reflect the suggestion before that. :bow:

impactaky: Sorry, for late reply :bow:
tensor.raw_data still needed, because onnx.external_data_helper.set_external_data require raw_data field
https://github.com/onnx/onnx/blob/1cc5e12f95c168146a912e3a6837d51f5d0f693b/onnx/external_data_helper.py#L83-L84

take-cheeze: Sorry I've forgot that!
Then this is resolved. Thank you for the explanation!

PR #368, opened 2021-09-29
Fix bug in train_step

HiroakiMikami:
`train_step` raises error (`TypeError: len() of a 0-d tensor `) when using amp and output is a scalar tensor.


emcastillo (approved):
Thank you!

PR #375, opened 2021-09-29
[onnx] Consider tensors using external data

disktnk:
If the target tensor uses `external_data`, cannot convert to numpy array when the external data is in the `location`. Object of strip function is to reduce ONNX size, so assume that the goal is accomplished. If valid `location` is set = invalid JSON is set, skip strip.

### DIFF
--- a/pytorch_pfn_extras/onnx/strip_large_tensor.py
+++ b/pytorch_pfn_extras/onnx/strip_large_tensor.py
@@ -27,7 +27,7 @@ def _is_stripped_or_set_external(tensor):
             return external_value_dict.get('type', '') == 'stripped'
         except ValueError:
             # Invalid JSON, indicating `external_data.value` contains
-            # a file path, assume the tensor has no value means already stripped
+            # a file path. Treat the tensor as if it is already stripped.
             return True
     return False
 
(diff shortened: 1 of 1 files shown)


okdshin:
I think `_is_stripped` should be renamed like `_is_stripped_or_external` reflecting the modification.

pytorch_pfn_extras/onnx/strip_large_tensor.py
```diff
@@ -27,8 +27,8 @@ def _is_stripped(tensor):
             return external_value_dict.get('type', '') == 'stripped'
         except ValueError:
             # Invalid JSON, indicating `external_data.value` contains
-            # a file path
-            continue
+            # a file path, assume the tensor has no value means already stripped
```
kmaehashi, line 30: This fixes flake8 (line length limit) error. Correct me if my understanding is wrong 😃 
```suggestion
            # a file path. Treat the tensor as if it is already stripped.
```

kmaehashi (changes requested):

pytorch_pfn_extras/onnx/strip_large_tensor.py
kmaehashi: As this is comment-only fix, let me add this fix so that we can include this PR in the release today. Feel free to let me know if you have any concern 😄

kmaehashi (approved):

PR #380, opened 2021-10-01
Remove py.typed

asi1024:


kmaehashi:
Thanks! Just want to confirm, does this fix work well with user's code e.g. using Lazy modules?
I think `setup.py` also needs to be updated.

kmaehashi:
As discussed offline, if we remove `py.typed` user code inheriting from PPE classes will start to raise a mypy error, so that's what we want to avoid.
Let's move on to annotate all public APIs in PPE and remove `mypy: ignore-errors`.

PR #382, opened 2021-10-03
Support `concurrency` option in `OutputComparer`

### LINKED ISSUE #354: OutputComparer: add API to limit `max_pool` size
It can be convenient when comparing against engines with limitation (e.g., memory size)

asi1024:
Close #354.

~TODO: docstring~

### DIFF
--- a/pytorch_pfn_extras/utils/comparer.py
+++ b/pytorch_pfn_extras/utils/comparer.py
@@ -93,7 +93,7 @@ class OutputsComparer:
     def __init__(
             self, engines, to_compare_keys=None, *,
             compare_fn=_default_comparer,
-            max_pool=None,
+            concurrency=None,
     ):
         """A class for comparison of iteration outputs.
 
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/utils/comparer.py
```diff
@@ -93,6 +93,7 @@ class OutputsComparer:
     def __init__(
             self, engines, to_compare_keys=None, *,
             compare_fn=_default_comparer,
+            max_pool=None,
```
kmaehashi, line 96: Maybe naming it `concurrency` is better?
```suggestion
            concurrency=None,
```

asi1024:
PTAL.

pytorch_pfn_extras/utils/comparer.py
```diff
@@ -105,6 +106,9 @@ def __init__(
                 A set of keys of output dict to compare.
             compare_fn (function):
                 Comparison function. Default is ``get_default_comparer()``.
+            concurrency (int, optional):
+                The upper bound limit on the number of workers that run concurrently.
```
emcastillo, line 110:
```suggestion
                Maximum number of workers that run concurrently.
```

Maybe this is a bit clearer? just a nit.

asi1024: Fixed. Thanks!

emcastillo (approved):

PR #387, opened 2021-10-05
Publicize `Trainer` and `Evaluator`

asi1024:


pytorch_pfn_extras/runtime/_runtime.py
```diff
@@ -11,8 +11,8 @@
 DeviceLike = Union[str, torch.device]
 
 if TYPE_CHECKING:
```
kmaehashi, line 13: Let's remove `if TYPE_CHECKING:` and replace `'Trainer'` with `Trainer` so that users can easily copy method signatures of base classes.

kmaehashi (changes requested):
Comment (as discussed offline)

pytorch_pfn_extras/runtime/_runtime.py
asi1024: Fixed. PTAL.

asi1024: FIxed.

kmaehashi (approved):
Thanks, LGTM!

PR #341, opened 2021-09-13
Reduce snapshot size when using `append=True` in `LogReport`

asi1024:
Fixes #245

PR #391, opened 2021-10-12
Remove deprecated `log_name=None` support and fix documentation error in LogReport

kmaehashi:
Depends on #341


kmaehashi:
Rebased!
/test

PR #421, opened 2021-11-10
Allow to return dict of `DeferredResult`

emcastillo:
In some cases, instead of returning an awaitable object that yields the dict of the results, we would like to return a dict with entries of awaitable objects.


emcastillo:
PTAL!

asi1024:
LGTM.

asi1024 (approved):

PR #423, opened 2021-11-10
Make comparer error message comfortable

asi1024:
```
AssertionError: Comparing 'cpu' and 'gpu' in 'output:a'
With rtol=1e-07 and atol=0, found 1 element(s) (out of 1) whose difference(s) exceeded the margin of error (including 0 nan comparisons). The greatest difference was 0.5 (1.0 vs. 0.5), which occurred at index 0.
```


emcastillo (approved):
LGTM!!

PR #433, opened 2021-11-24
Call `writer.finalize()` in extensions that use the writer

emcastillo:
In some writers, if close is not called before interpreter shutdown, lots of nasty things can happen as the interpreter is in an undefined state so we can see non-reproducible hangs when trying to join subprocesses, etc.

This PR assures that extensions using a writer other than the manager close it when the training loop is finished.

### DIFF
--- a/pytorch_pfn_extras/training/extensions/log_report.py
+++ b/pytorch_pfn_extras/training/extensions/log_report.py
@@ -263,5 +263,5 @@ class LogReport(extension.Extension):
         return pandas.DataFrame(self._log_looker.get())
 
     def finalize(self) -> None:
-        if self._writer is not None:
+        if self._writer is not None and hasattr(self._writer, 'finalize'):
             self._writer.finalize()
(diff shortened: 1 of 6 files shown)


pytorch_pfn_extras/training/extensions/log_report.py
```diff
@@ -253,3 +253,7 @@ def to_dataframe(self):
                 "Need to install pandas to use `to_dataframe` method."
             )
         return pandas.DataFrame(self._log_looker.get())
+
+    def finalize(self) -> None:
+        if self._writer is not None and hasattr(self._writer, 'finalize'):
```
kmaehashi, line 258: As `finalize` is the interface defined in the base writer class, let's call it directly rather than checking with hasattr.

kmaehashi (changes requested):
Thanks!

pytorch_pfn_extras/training/extensions/log_report.py
emcastillo: good catch, this was my mistake as I just copied the function straight from the snapshot extension. Fixed now! 🙇‍♂️

kmaehashi:
#439 needs to be merged first , as `writer` in each extension instance may point to the same Writer instance.

kmaehashi (approved):

PR #438, opened 2021-11-25
Comparer: Enhance error message from compare_targets

### LINKED ISSUE #468: Comparer: show results aggregated by iteration
Currently, comparer fails immediately when any of the differences are found. However, it is difficult to find the root cause as the comparison may not run in the order of computation. It is better to report errors aggregated by iteration.

asi1024:
Close #468.


asi1024:
@kmaehashi PTAL!

kmaehashi (approved):

Issue #336, opened 2021-09-13
Remove `# mypy: ignore-errors`


kmaehashi:

kmaehashi:
These are high priority and would like to include this in the next release;
```
pytorch_pfn_extras/_tensor.py:# mypy: ignore-errors
pytorch_pfn_extras/profiler/_time_summary.py:# mypy: ignore-errors
pytorch_pfn_extras/runtime/_runtime.py:# mypy: ignore-errors
pytorch_pfn_extras/training/_trainer.py:# mypy: ignore-errors
pytorch_pfn_extras/training/triggers/early_stopping_trigger.py:# mypy: ignore-errors
pytorch_pfn_extras/training/triggers/manual_schedule_trigger.py:# mypy: ignore-errors
pytorch_pfn_extras/training/triggers/minmax_value_trigger.py:# mypy: ignore-errors
pytorch_pfn_extras/training/triggers/once_trigger.py:# mypy: ignore-errors
pytorch_pfn_extras/utils/checkpoint.py:# mypy: ignore-errors
```

kmaehashi:
The remainings are comparer (#480) and dataset APIs. As the number of dataset API users is not so large we are changing the priority to low.

PR #450, opened 2021-12-01
Typing: utils.checkpoint

asi1024:
Part of #336.

### DIFF
--- a/pytorch_pfn_extras/utils/checkpoint.py
+++ b/pytorch_pfn_extras/utils/checkpoint.py
@@ -1,5 +1,6 @@
-# mypy: ignore-errors
+from typing import Any
 
+import torch.nn
 import torch.utils.checkpoint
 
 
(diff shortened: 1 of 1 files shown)


pytorch_pfn_extras/utils/checkpoint.py
```diff
@@ -16,13 +17,18 @@ class _CheckpointFunction(torch.utils.checkpoint.CheckpointFunction):
     persistent parameters.
     """
     @staticmethod
-    def forward(ctx, run_function, preserve_rng_state, *args):
+    def forward(  # type: ignore[override]
+            ctx: Any,
+            run_function: torch.nn.Module,
```
kmaehashi, line 22: This may cause a problem once type annotation is added `torch.utils.checkpoint.CheckpointFunction`, which expects a callable.

```suggestion
            run_function: Any,
```

kmaehashi (approved):

PR #470, opened 2021-12-22
Fix invalid operator_export_type

asi1024:


take-cheeze (approved):
Seems this path isn't tested.
LGTM 👍 
https://github.com/pytorch/pytorch/blob/c76c6e9bd345ba697ea2a6fb8c2e24f051bb2e00/torch/csrc/onnx/init.cpp#L32

PR #471, opened 2021-12-22
Typing: ppe.onnx.export_testcase

asi1024:
~Blocked by #470.~

### DIFF
--- a/pytorch_pfn_extras/onnx/export_testcase.py
+++ b/pytorch_pfn_extras/onnx/export_testcase.py
@@ -1,11 +1,10 @@
-# mypy: ignore-errors
-
 import datetime
 import io
 import itertools
 import json
 import os
 import subprocess
+from typing import Any, Dict, IO, Mapping, Optional, Sequence, Tuple, Union
 import warnings
 
 import onnx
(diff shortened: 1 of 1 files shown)


asi1024:
Rebased.

pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -79,7 +87,12 @@ def strip_cmd(cmd):
     return ret
 
 
-def _export_util(model, args, f, **kwargs):
+def _export_util(
+        model: torch.nn.Module,
+        args: Tuple[Any, ...],
```
take-cheeze, line 92:
```suggestion
        args: Sequence[Any, ...],
```

take-cheeze: For compatibility with [torch.onnx.export](https://pytorch.org/docs/stable/onnx.html#torch.onnx.export) maybe `Union[Sequence[Any, ...], torch.Tensor]` might be better.
Sometimes list would be also passed so allowing it with `Sequence` would be better

asi1024: Fixed in 5058f03! Thanks!

asi1024:
Rebased.

take-cheeze (approved):
👍

PR #474, opened 2021-12-23
[WIP] Check torch.distributed.is_available() before using ProcessGroup type

take-cheeze:


take-cheeze:
https://github.com/pfnet/pytorch-pfn-extras/runs/4615623502?check_suite_focus=true#step:5:76

take-cheeze:
This seems to be difficult to handle in mypy so I'll close

PR #477, opened 2021-12-24
[onnx] Unstrip script

disktnk:
In order to run with onnx runtimes even if the target ONNX are stripped.

```
$ python -m pytorch_pfn_extras.onnx.unstrip_tensor target.onnx
$ python -m pytorch_pfn_extras.onnx.unstrip_tensor target_dir/
```

TODO
* [x] support `*.onnx`
* [x] support `*.pb`
* [x] support directory, unstrip `model.onnx`, input/output at once
* [x] unit test
* [x] docs

### DIFF
--- a/pytorch_pfn_extras/onnx/unstrip_tensor.py
+++ b/pytorch_pfn_extras/onnx/unstrip_tensor.py
@@ -66,6 +66,16 @@ def _unstrip_tensor_from_path(path: Path) -> onnx.TensorProto:
 
 
 def unstrip(path: str, out_path: str = "") -> None:
+    """Unstrip ONNX models and test data(.pb).
+
+    Add tensor(raw data) to the target ONNXs (and test data).
+    Values are random following mean and variance written in meta information.
+
+    Args:
+        path (str): The target directory path, ONNX file, or Tensor (Protobuf)
+            file path.
+        out_path (str): Output path to be written.
+    """
     target_path = Path(path)
     if not target_path.exists():
         print(f"Error: the target path is not found, {path}")
--- a/pytorch_pfn_extras/onnx/unstrip_tensor.py
+++ b/pytorch_pfn_extras/onnx/unstrip_tensor.py
@@ -1,6 +1,7 @@
 import argparse
 import json
 import math
+from pathlib import Path
 
 import onnx
 import onnx.numpy_helper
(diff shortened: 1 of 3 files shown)


pytorch_pfn_extras/onnx/unstrip_tensor.py
```diff
@@ -0,0 +1,131 @@
...
+def _unstrip_graph(graph: onnx.GraphProto) -> None:
+    for init in graph.initializer:
+        if not strip._is_stripped_or_set_external(init):
+            continue
+        _unstrip_tensor(init)
+
+    for node in graph.node:
+        for attr in node.attribute:
+            if attr.HasField("g"):
+                _unstrip_graph(attr.g)
```
take-cheeze, line 51: I've thought of also un-stripping tensor attributes like following:
```suggestion
            if attr.HasField("g"):
                _unstrip_graph(attr.g)
            elif attr.HasField("t"):
                _unstrip_tensor(attr.t)
```
But current strip doesn't do it so never mind
https://github.com/pfnet/pytorch-pfn-extras/blob/785cc23138a0a648a83dc90c4f09302dc9f0938d/pytorch_pfn_extras/onnx/strip_large_tensor.py#L59

pytorch_pfn_extras/onnx/unstrip_tensor.py
```diff
@@ -0,0 +1,131 @@
...
+            file path.
+        out_path (str): Output path to be written.
+    """
+    target_path = Path(path)
+    if not target_path.exists():
+        print(f"Error: the target path is not found, {path}")
+        return None
+    written_path = Path(out_path if out_path else path)
+
+    if target_path.is_dir():
```
take-cheeze, line 85: https://github.com/disktnk/pytorch-pfn-extras/pull/1 meta.json should be also updated

disktnk: supported to handle `meta.json` if existed

pytorch_pfn_extras/onnx/unstrip_tensor.py
disktnk: it's better to support `attr.t`. will try dealing with another PR along with `strip` method.

take-cheeze (approved):
LGTM 👍

kmaehashi (approved):

PR #478, opened 2021-12-28
[ONNX] Skip metadata export in test if possible

take-cheeze:
Metadata export is quite expensive since they call git command

### DIFF
--- a/tests/pytorch_pfn_extras_tests/onnx_tests/test_export_testcase.py
+++ b/tests/pytorch_pfn_extras_tests/onnx_tests/test_export_testcase.py
@@ -70,7 +70,7 @@ def test_export_testcase():
     model = Net().to('cpu')
     x = torch.zeros((1, 1, 28, 28))
 
-    output_dir = _helper(model, x, 'mnist', output_grad=True)
+    output_dir = _helper(model, x, 'mnist', output_grad=True, metadata=True)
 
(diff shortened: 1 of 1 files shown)


tests/pytorch_pfn_extras_tests/onnx_tests/test_export_testcase.py
```diff
@@ -254,7 +256,8 @@ def test_export_testcase_strip_large_tensor_data():
 
     output_dir = _helper(
         model, x, 'mnist_stripped_tensor_data',
-        output_grad=True, strip_large_tensor_data=True)
+        output_grad=True, strip_large_tensor_data=True,
+        metadata=True)
```
emcastillo, line 260:
```suggestion
        metadata=False)
```
Sorry, did you mean this? 😅

take-cheeze: This test is the only test that needed metadata so I'm enabling it explicitly
(Striped test case will have metadata to tell it. Though this needs to be warned when `metadata=False` and `strip_large_tensor_data=True` are both enabled)

emcastillo: ok! I understood the contrary hahaha. Thanks!

emcastillo (approved):

PR #484, opened 2022-01-05
[ONNX] Add pure python torch.onnx.export implementation

take-cheeze:
- Adds `use_pfto=False` option to switch to new exproter
- Custom stubs is added for `torch._C` (Graph, Block, Node, Value, Type classes) module for torch IR interfaces
- Instead of internal API `torch.jit.trace` method is used to convert model to torch jit model
- `onnxruntime` will be used to test models
- `torchvision` will be used to test real world models

TODOs before merge
- [x] Use version checker from https://github.com/pfnet/pytorch-pfn-extras/pull/501
- [x] Catch up to latest master branch
- [x] Updates for reviews
- [x] Install packages for testing: onnxruntime, torchvision


stubs/torch/_C/__init__.pyi
```diff
@@ -0,0 +1,2416 @@
...
+
+# Defined in torch/csrc/jit/python/python_ir.cpp
+# Not actually defined in python_ir.cpp, not sure where they are.
+class IValue:
+    ...
+Stack = List[IValue]
+
+
+# Diff(twata): Renamed JitType to Type, added methods too
+class Type:
```
take-cheeze: `s/JitType/Type/`

stubs/torch/_C/__init__.pyi
```diff
@@ -0,0 +1,2416 @@
...
+    def scopeName(self) -> str: ...
+    def sourceRange(self) -> str: ...
+    def __getitem__(self, attr: str) -> Any: ...
+    def blocks(self) -> Iterator[Block]: ...
+    def pyobj(self) -> Any: ...
+    def scalar_args(self) -> List[Any]: ...
+    def moveBefore(self, n: Node) -> None: ...
+    def mustBeNone(self) -> bool: ...
+    def addInput(self, v: Value) -> None: ...
+    ...
```
take-cheeze: Methods added to use in pfto

pytorch_pfn_extras/onnx/export/export.py
```diff
@@ -0,0 +1,879 @@
+import dataclasses
```
emcastillo, line 1: I feel that this should be an independent python-package :)
ppe-onnx?

take-cheeze: Yes it could be a independent package.
Though all other `ppe.onnx` features should also be in it for easier integration test

kmaehashi: That could be possible, but I still think it's better to have it in PPE, because separation means:
* Forcing users to install two separate libraries
* Add extra overhead to take care of version compatibility between PPE and PPE-ONNX as they are relatively tightly-coupled

PPE has a flexible release policy, so in case we need a hot-fix inside these new modules, we can just cut a release 😄

kmaehashi: Btw, `pytorch_pfn_extras/onnx/export` module shadows the existing public API `pytorch_pfn_extras.onnx.export` (which is defined at `pytorch_pfn_extras.onnx.export_testcase.export`) so the direcotry needs to be renamed. Maybe `pytorch_pfn_extras/onnx/pfto_exporter` or something?

kmaehashi:
@emcastillo and I checked the PR and this looks great! :+1::+1: Could you continue completing the remaining work?

kmaehashi:
Please retrigger tests after this job completes! https://ci.preferred.jp/pytorch-pfn-extras.prep-linux/93215/

kmaehashi:
Regenerated: https://ci.preferred.jp/r/job/93693

/test

take-cheeze:
@kmaehashi Thank you for your help. Now CI jobs are all success!

kmaehashi (approved):

emcastillo (approved):
LGTM!

asi1024 (approved):

PR #486, opened 2022-01-05
Change `out_dir` to better reflect the fs needs

### LINKED ISSUE #475: [RFC] don't pass `out_dir` to writers
After an offline discussion with @belltailjp,
he proposes that the path to save objects through the writers should be stored in the filesystem object instead of passing an`out_dir` string directly to the writers.

I agree with this proposal

emcastillo:
Closes #475 

~`out_dir` has now a default value of '.' since using an empty one required major changes.~

I am not happy at all of how this PR turn out to be, I think we may need to rethink how fs and out_dir interact

### DIFF
--- a/pytorch_pfn_extras/writing/_writer_base.py
+++ b/pytorch_pfn_extras/writing/_writer_base.py
@@ -40,7 +40,7 @@ class _PosixFileSystem(object):
 
     This class currently abstracts POSIX
     """
-    def __init__(self, root='/') -> None:
+    def __init__(self, root: str = '/') -> None:
         self._root = root
 
     def get_actual_path(self, path: str) -> str:
@@ -185,7 +185,7 @@ class Writer:
     def __init__(
             self,
             fs: _FileSystem = None,
-            out_dir: str = '',
+            out_dir: str = '.',
     ) -> None:
         self._post_save_hooks: List[_HookFun] = []
         self.fs = fs or _PosixFileSystem()
@@ -201,7 +201,7 @@ class Writer:
             savefun: Optional[_SaveFun] = None,
             append: bool = False
     ) -> None:
-        """Invokes the actual snapshot function.
+        """Does the actual writing to the file.
 
         This method is invoked by a
         :class:`~pytorch_pfn_extras.training.extensions.Snapshot` object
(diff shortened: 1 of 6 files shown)


emcastillo:
I think it looks a bit better now, but still not 100% happy

@belltailjp , do you have any suggestions 😇 ?

belltailjp:
Thanks so much for the quick prototype!!
This branch worked well for me as expected.

> I think we may need to rethink how fs and out_dir interact

Right... I feel that the responsibility to handle where to save things into is currently a bit diverged (`fs` vs `out_dir` to `Writer.__init__` and `Writer.save`). IMHO the ideal situation would be to integrate them into the `fs` mechanism, so Writer (and extensions) can and should only concern about sub-directories visible through the `fs`, but I agree that it's gonna be a little too much breaking change. In addition doing so would force user to always make FS.
```python
writer = SimpleWriter(PosixFileSystem("/path/to/some/dir"))
```

pytorch_pfn_extras/writing/_writer_base.py
```diff
@@ -89,6 +93,7 @@ def list(
             if path_or_prefix is None:
                 raise ValueError(
                     "'path_or_prefix' must not be None in recursive mode.")
+            path_or_prefix = self.get_actual_path(path_or_prefix)
```
kmaehashi, line 96: Need to normalize path even if the `recursive=False` case.

pytorch_pfn_extras/writing/_writer_base.py
```diff
@@ -140,9 +149,11 @@ def makedirs(
...
             exist_ok: bool = False
     ) -> None:
+        file_path = self.get_actual_path(file_path)
         return os.makedirs(file_path, mode, exist_ok)
 
     def exists(self, file_path: str) -> bool:
+        file_path = self.get_actual_path(file_path)
         return os.path.exists(file_path)
 
     def rename(self, src: str, dst: str) -> None:
```
kmaehashi, line 159: `get_actual_path` needed?

kmaehashi (changes requested):

pytorch_pfn_extras/writing/_writer_base.py
emcastillo: fixed!

pytorch_pfn_extras/writing/_writer_base.py
emcastillo: good catch, old code

emcastillo:
PTAL 😄

pytorch_pfn_extras/writing/_writer_base.py
```diff
@@ -89,11 +89,11 @@ def list(
             path_or_prefix: Optional[str] = None,
             recursive: bool = False,
     ) -> Iterator[str]:
+        path_or_prefix = self.get_actual_path(path_or_prefix)
```
kmaehashi: `get_actual_path` could be None here 😅

emcastillo: yeah, I think this should fix it, `get_actual_path` fails and `os.scandir` admits `None`

kmaehashi (approved):

Issue #398, opened 2021-10-20
ExtensionsManager should not call extensions when the iteration fail


kmaehashi:
https://github.com/pfnet/pytorch-pfn-extras/blob/master/pytorch_pfn_extras/training/manager.py#L634-L648

PR #490, opened 2022-01-07
Fix ExtensionsManager to call Extension.on_error

### LINKED ISSUE #397: ExtensionsManager need to call of `Extension.on_error` handler on error
Snapshot extensions provides `snapshot_on_error` option, but currently this is not working as `on_error` handler is not called from the manager.

asi1024:
Close #397 and #398.

### DIFF
--- a/pytorch_pfn_extras/training/manager.py
+++ b/pytorch_pfn_extras/training/manager.py
@@ -160,6 +160,7 @@ class _BaseExtensionsManager:
         self.reporter = reporting.Reporter()
         self._transform_model = transform_model
         self._start_extensions_called = False
+        self._run_on_error_called = False
 
         # Indicates whether models can be accessed from extensions in the
         # current iteration.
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/training/manager.py
```diff
@@ -435,6 +436,14 @@ def get_extension(self, name: str) -> extension_module.Extension:
         else:
             raise ValueError('extension %s not found' % name)
 
+    def _run_on_error(self, exc: Exception) -> None:
+        if not self._run_on_error_called:
+            self._run_on_error_called = True
+            tb = exc.__traceback__
+            assert tb is not None
+            for _, entry in self.extensions:
+                entry.extension.on_error(self, exc, tb)
```
emcastillo, line 445: I think we need to check if the extension has `on_error` method.

emcastillo (approved):
LGTM!

pytorch_pfn_extras/training/manager.py
emcastillo: discussed offline, `entry` is always wrapped so `on_error` is guaranteed to be present.

PR #498, opened 2022-01-21
Disable profiler by default in trainer/manager (no-compat)

emcastillo:
The `TimeSumary` object causes problems of interop with multiprocessing because of the destruction order of threads and queues, resulting in deadlocks under some python/os configurations.

Since this feature is not widely used, we set it to `False` by default and make users to explicitly request it.

Confirmed that #496 was fixed


kmaehashi (approved):

kmaehashi:
will trigger failling tests again after docker image rebuild: https://ci.preferred.jp/pytorch-pfn-extras.prep-linux/92237/

PR #499, opened 2022-01-21
Fix filename format of `comparer.dump(evaluator)`

asi1024:
Fix filename duplication.


kmaehashi:
Discussed offline, need to think about the cases:
* multiple evaluators
* single evaluator but multiple evaluations in the same epoch

emcastillo (approved):

PR #501, opened 2022-01-25
Add version check function for special torch package version names

take-cheeze:
Rel https://github.com/pfnet/pytorch-pfn-extras/pull/484
When using torch with version names like `1.9.0+cpu+clang10`, `version.Version` will fail to parse it so this would be a workaround for special package version names

### DIFF
--- a/pytorch_pfn_extras/torch_version.py
+++ b/pytorch_pfn_extras/torch_version.py
@@ -1,9 +1,0 @@
-import torch
-from packaging import version
-
-
-torch_version = version.Version(torch.__version__.split("+")[0])
-
-
-def requires(version: str, *, package='torch') -> bool:
-    return torch_version >= version.Version(ver)
--- a/pytorch_pfn_extras/torch_version.py
+++ b/pytorch_pfn_extras/torch_version.py
@@ -1,0 +1,9 @@
+import torch
+from packaging import version
+
+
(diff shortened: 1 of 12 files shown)


pytorch_pfn_extras/torch_version.py
```diff
@@ -0,0 +1,9 @@
+import torch
```
kmaehashi, line 1: Please prefix the filename with `_` for private modules.

pytorch_pfn_extras/torch_version.py
```diff
@@ -0,0 +1,9 @@
...
+from packaging import version
+
+
+torch_version = version.Version(torch.__version__.split("+")[0])
+
+
+def is_available(ver: str) -> bool:
```
kmaehashi, line 8: How about defining it as 

```suggestion
def requires(version: str, *, package='torch') -> bool:
```

and get version via `pkg_resources.get_distribution(package).version` so that this function can be used for family packages such as `torchvision` as well?

kmaehashi (commented):
Thanks!

kmaehashi (dismissed):
LGTM!

take-cheeze:
https://ci.preferred.jp/pytorch-pfn-extras.torch110-win/92604/

kmaehashi (approved):

PR #516, opened 2022-02-15
Fix `UnboundLocalError` in pfto exporter

asi1024:
If `self.input_names is not None` is evaluated to `False`, the following error is raised.
```
UnboundLocalError: local variable 'input_names' referenced before assignment
```


take-cheeze (approved):
Sorry for latency. Thank you!!! LGTM 👍

PR #518, opened 2022-02-19
Best observation extension

belltailjp:
I'd like to propose `BestValue`, `MaxValue` and `MinValue` extensions which traces the best/max/min value and timing (iter/epoch) by checking an internal observation every time their base trigger is ignited.

The usage I consider particularly is to combine with Optuna; may not be always but sometimes we want to report the best accuracy (or whatever score) which appeared *in the middle of training* rather than the *last one after training completed*, in order to prevent a risk of reporting an overfit score.

```python
def objective(trial):
    ...
    manager = ExtensionsManager(...)
    max_score = ppe.training.extension.MaxValue("validation/main/accuracy")
    manager.extend(max_score, trigger=(1, 'epoch'))
    ...
    # run training for several epochs
    ...
    return max_score.best_value
    # return manager.observations["validation/main/accuracy"]    # rather than this
```

There already are the [`Best/Min/MaxValueTrigger`](https://github.com/pfnet/pytorch-pfn-extras/blob/master/pytorch_pfn_extras/training/triggers/minmax_value_trigger.py) which basically do the same thing but work as a trigger. Actually what I propose here is essentially an extension-version of them.
Maybe adding a few interfaces to such trigger classes instead of adding new extensions sounds to be sufficient, but being extensions has several additional benefits depending on the use-cases:
- An extension can be accessed through an Extension Manager (`get_extension(name)`); no need to pass the extension object around.
- When we do nothing more than just recording best values, with only triggers we may need a bit odd ways; manually call `best_value_trigger(manager)` or create an empty extension with the trigger.


pytorch_pfn_extras/training/extensions/best_value.py
```diff
@@ -0,0 +1,145 @@
...
+        if not self._interval_trigger(manager):
+            return
+
+        stats = summary.compute_mean()
+        if key not in stats:
+            raise RuntimeError('Key "{}" not found in the observation '
+                               '(current available keys are {})'
+                               .format(key, list(stats.keys())))
+
+        value = float(stats[key])  # copy to CPU
```
emcastillo, line 60: Can this copy be avoided? :)

belltailjp: Fixed by re-using the `BestValueTrigger` internally!

Note: The small change applied in the BestValueTrigger [here](https://github.com/pfnet/pytorch-pfn-extras/pull/518/files#diff-5881bfe2983a95132b388de54eb31d5108a5bad3511c6e39138afcbaccdf331e) which is brought from my original code doesn't actually change the behavior, but just changes (enriches) the message in exceptions.

emcastillo: `BestValue` syncs the cpu & gpu so if you call it multiple times inside an epoch, performance is likely going to suffer a lot.

Would it be possible to just do `torch.max(prev_value, new_value)` and do not use this inside any `if`?
I am not sure if this is possible, but what this aims is to have the best value at the end of the epoch calculated in the device without interrupting the stream.

Thanks!

belltailjp: Aha, I see... I misunderstood that you meant *copy* by copying-and-pasting the core code from `BestValueTrigger`... I think that DRY would still be good so maybe I don't need to revert it, though.

BTW, the performance concern is not what is introduced by this PR (`BestValueTrigger` already has it), so may I leave it out of scope in this PR?
I'm also wondering a few points on introducing the suggested way, not sure they get the point though:
- In case the best value trigger object is loaded from a state dict, its internals are on CPU (`float`) at that moment, so comparison using `torch.{min,max}` would make things a little complicated to handle type/device inconsistency.
- Interfaces to get the best epoch/iters are sometimes useful; in case we are interested in not only the best values but also the actual model file if we save models e.g., every epochs. I guess `torch.max` approach cannot simply achieve this.

emcastillo: no worries!

emcastillo: FYI If you want to access the model parameters from the extension you need to set
`needs_model_state=True`

emcastillo (approved):

pytorch_pfn_extras/training/extensions/best_value.py
belltailjp: Thanks for the comment as well as reviewing!
I wasn't aware of `needs_model_state=True` option. "corresponding model file" meant a little more stupid idea; one can easily get the filename of the best model file by `"model_epoch_{best_value_ext.best_epoch}"`

PR #522, opened 2022-03-04
Test with CPU in GH actions

### LINKED ISSUE #510: Introduce CPU-only tests in GitHub Actions
Now that markers are added in #509, we can selectively run CPU tests.

emcastillo:
Closes #510 

Discuss: Python versions

### DIFF
--- a/.github/workflows/test-cpu.yml
+++ b/.github/workflows/test-cpu.yml
@@ -5,15 +5,6 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-18.04
-    strategy:
-      matrix:
-        include:
-           - torch: '1.8.*'
-             torchvision: '0.9.*'
-           - torch: '1.9.*'
-             torchvision: '0.10.*'
-           - torch: '1.10.*'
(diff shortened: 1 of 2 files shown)


.github/workflows/test-cpu.yml
```diff
@@ -0,0 +1,42 @@
...
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        include:
+           - torch: '1.8.*'
+             torchvision: '0.9.*'
```
emcastillo, line 12: Should we set python versions in the matrix, I think yes :)

kmaehashi:
I rather think only one variant (fixed PyTorch & Python version) is enough because (1) CPU-only test is to confirm that test markers are correctly specified to each test (2) We will run FlexCI for all PRs which runs every test (3) Slots of GitHub Actions is shared within the OSS products under the `pfnet` organization so I don't want to monopolize it 😅

emcastillo:
only one test now :)

kmaehashi (approved):
LGTM!

PR #524, opened 2022-03-04
Fix version check for versions with '-' splitter

take-cheeze:
For example `torch.__version__` returns `1.9.0+cu102` but in some env `pkg_resources.get_distribution(package).version` returns `1.9.0-cu102`


emcastillo (approved):
Thank you!

PR #531, opened 2022-03-11
pfto: Fix onnx export of torch.cat

take-cheeze:
Previous: `onnx::ListConstruct` (prim::ListConstruct) -> `onnx::ConcatFromSequence` (aten::cat)
Now: `onnx::Concat` only


asi1024 (approved):

asi1024:
@take-cheeze Could you check the test failure?

take-cheeze:
Seems infinite loop is happening in ppe.onnx test. Will try reproduce on my env

asi1024 (approved):

PR #535, opened 2022-03-31
Use `ManagerProxy` in extensions

emcastillo:
It was only being used for the trigger.

Adds methods to retrieve the real iteration/execution in case they are needed


emcastillo:
Lets favor #536 over this

PR #538, opened 2022-04-08
Support `dataclass` for batches

### LINKED ISSUE #537: Will the default runtime support dataclass?
I'm trying to use the trainer with batches as `dataclass` but I found it only supports `tupple` which has attribute `_fields`, `dict`, `list`, `tupple`, and `torch.Tensor`.
https://github.com/pfnet/pytorch-pfn-extras/blob/0c16c95ed97d7e76edd1db6ba25ddda2042d4440/pytorch_pfn_extras/runtime/_runtime.py#L72-L84

I guess it just follows PyTorch's default collate function.
https://github.com/pytorch/pytorch/blob/9c56cc475523fb2b7cd6131c9b8eb2ab87724cac/torch/utils/data/_utils/collate.py#L50-L74

On the other hand, `dataclass` is becoming much more common as data structure these days and I would like ppe to support it as input batch class. I know it's possible to support dataclass using `register` to add a new runtime. So my question is, do you have a plan to support `dataclass` by built-in runtime?

emcastillo:
Closes #537

### DIFF
--- a/pytorch_pfn_extras/runtime/_runtime.py
+++ b/pytorch_pfn_extras/runtime/_runtime.py
@@ -1,4 +1,5 @@
 import contextlib
+import dataclasses
 
 from typing import Any, Dict, Generator, Iterable, Optional, Tuple, Union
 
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/runtime/_runtime.py
```diff
@@ -73,14 +74,16 @@ def convert_batch(self, args: Any) -> Any:
...
-        if isinstance(args, (list, tuple)):
+        elif isinstance(args, (list, tuple)):
             return [
                 self.move_tensor(v) if isinstance(v, torch.Tensor) else v
                 for v in args
             ]
-        if isinstance(args, torch.Tensor):
+        elif isinstance(args, torch.Tensor):
             return self.move_tensor(args)
         return args
```
asi1024, line 88: For the consistency of fixes.
```suggestion
        else:
            return args
```

PR #540, opened 2022-04-11
Ensure that extensions are finalized in manager

emcastillo:
Call `finalize` from trainer too in case that manager didn't do it.


kmaehashi:
Closing as #536 got merged!

PR #543, opened 2022-05-09
Add `update_via_args` to config

emcastillo:
Allow updating the config system via parameters that can be specified in the cmdline

```python
config = pytorch_pfn_extras.config.Config.load_path(json_path)
config.update_via_args([parameter.split("=") for parameter in args.parameters])

# Run script
python script.py --parameters model/layer=2 dataloader/batchsize=16
```

Thanks to @dhgrs !

### DIFF
--- a/pytorch_pfn_extras/config.py
+++ b/pytorch_pfn_extras/config.py
@@ -138,7 +138,7 @@ class Config(object):
             else:
                 return cache(config)
 
-    def update_via_args(self, args: List[Tuple[str, Any]]) -> None:
+    def update_via_args(self, args: Sequence[Tuple[str, Any]]) -> None:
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/config.py
```diff
@@ -138,6 +138,11 @@ def cache(value: Any) -> Any:
             else:
                 return cache(config)
 
+    def update_via_args(self, args: List[Tuple[str, Any]]) -> None:
```
asi1024, line 141: How about using `typing.Sequence` instead?
```suggestion
    def update_via_args(self, args: Sequence[Tuple[str, Any]]) -> None:
```

asi1024 (approved):

PR #546, opened 2022-05-11
Finalize extension when expcetion is raised

HiroakiMikami:
finalize extensions and writer when user code raises expcetion in `run_iteration`.


pytorch_pfn_extras/training/manager.py
```diff
@@ -589,8 +589,8 @@ def run_iteration(
         if step_optimizers is not None:
             step_optimizers_names = step_optimizers
         self.observation = {}
-        with self.reporter.scope(self.observation):
-            try:
+        try:
+            with self.reporter.scope(self.observation):
```
kmaehashi, line 593: Could you elaborate on the motivation for the change?

kmaehashi:
Moving to #581

PR #547, opened 2022-05-12
Fix the type annotation of LogReport.log

zaburo-ch:
The type annotation in LogReport.log seems to be one of the following.
 - List[Any]
 - List[Mapping[str, Any]]
 - List[Dict[str, Any ]]
 - Sequence[Mapping[str, Any]]

I chose List[Mapping[str, Any]], but would like to know if something different is better.


asi1024 (approved):

asi1024:
LGTM!

PR #550, opened 2022-05-18
Respect types of previous items when setting config via args

emcastillo:
If updating the config items via arg.params, the user will need to convert the items to the datatypes increasing the burden.
This casts the new items to the type of the existing items in the config if any.


asi1024:
・てst

asi1024 (approved):

PR #551, opened 2022-05-18
PoC implementation for distributed evaluators

belltailjp:
Current evaluator mechanism (ppe.training.extensions.Evaluator) is not distributed-aware. Those who use distributed training currently run evaluation for the entire dataset on a single (typically rank=0) worker and let other workers just wait, or  maybe using only part of the evaluation dataset for evaluation during training.
The problem is two-fold.
- ppe Evaluator and Summary/DictSummary don't have mechanisms to collect evaluation results from other MPI workers.
- DistributedSampler is not suitable for evaluation.

This PoC provides several new interfaces for distributed evaluation on top of the ExtensionManager mechanism, by basically extends the Evaluator extension, which also includes:
* A mechanism to collect evaluation results from all workers (`__add__` operator for Summary and DictSummary).
* A variant of DistributedSampler with no sample duplication, which is dedicated for evaluation purpose
    * It guarantees each sample in a dataset is evaluated exactly once.
    * Batch-size won't be guaranteed to be fixed, number of batches for each worker either. So it should never be used for models that includes MPI communication.

Currently this doesn't include any implementations for `ppe.trainer`.
This is mainly because I first want to confirm if this approach is good to go or not. Also I have no idea how `create_evaluator` should look like to support distributed evaluation.

### DIFF
--- a/pytorch_pfn_extras/reporting.py
+++ b/pytorch_pfn_extras/reporting.py
@@ -367,6 +367,14 @@ class Summary:
         self._x2 = float(_nograd(to_load['_x2']))
         self._n = int(_nograd(to_load['_n']))
 
+    def __add__(self, other: "Summary") -> "Summary":
+        s = Summary()
+        s._x = self._x + other._x
+        s._x2 = self._x2 + other._x2
+        s._n = self._n + other._n
+        s._deferred = self._deferred + other._deferred
+        return s
+
 
 class DictSummary:
 
@@ -443,3 +451,16 @@ class DictSummary:
         self._summaries.clear()
         for name, summ_state in to_load.items():
             self._summaries[name].load_state_dict(summ_state)
+
+    def __add__(self, other: "DictSummary") -> "DictSummary":
+        ours, other = self._summaries, other._summaries
+        ds = DictSummary()
+        for k in sorted(list(set([*ours.keys(), *other.keys()]))):
+            if k not in ours:
+                s: Summary = other[k]
+            elif k not in other:
+                s: Summary = ours[k]
+            else:
+                s: Summary = ours[k] + other[k]
+            ds._summaries[k] = s
+        return ds
--- a/pytorch_pfn_extras/reporting.py
+++ b/pytorch_pfn_extras/reporting.py
@@ -453,14 +453,13 @@ class DictSummary:
             self._summaries[name].load_state_dict(summ_state)
 
     def __add__(self, other: "DictSummary") -> "DictSummary":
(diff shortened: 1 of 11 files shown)


pytorch_pfn_extras/reporting.py
```diff
@@ -367,6 +367,14 @@ def load_state_dict(self, to_load: Dict[str, Any]) -> None:
         self._x2 = float(_nograd(to_load['_x2']))
         self._n = int(_nograd(to_load['_n']))
 
+    def __add__(self, other: "Summary") -> "Summary":
+        s = Summary()
+        s._x = self._x + other._x
+        s._x2 = self._x2 + other._x2
+        s._n = self._n + other._n
+        s._deferred = self._deferred + other._deferred
```
emcastillo, line 375: I am not sure if this will work, we should probably synchronize all the deferred results before merging summaries.

pytorch_pfn_extras/training/_evaluator.py
```diff
@@ -151,8 +155,29 @@ def run(
...
+            metrics: Optional[Sequence['MetricType']] = None,
+    ):
+        super().__init__(handler, models, progress_bar=progress_bar, metrics=metrics)
+        if not torch.distributed.is_initialized():  # type: ignore[no-untyped-call]
+            raise RuntimeError("PyTorch distributed module is not initialized.")
+
+    def _gather_summaries(self) -> None:
+        world_size = torch.distributed.get_world_size()  # type: ignore[no-untyped-call]
+        summaries = [reporting.DictSummary() for _ in range(world_size)]
+        torch.distributed.all_gather_object(summaries, self._summary)  # type: ignore[no-untyped-call]
```
emcastillo, line 182: I am wondering if instead of getting the summaries, what would be the disadvantages of specifying a custom `key` that is in all summaries and perform an AllReduce op based in those?

```python
eval = DistributedEvaluator(gather_keys=["loss", "accuracy"])

class DistributedEvaluator(Evaluator):
    def _gather_summaries(self) -> None:
         for k in self._gather_keys:
             self._summary[k] = torch.distributed.all_reduce(self._summary[k], ...)
```
The problem I see with this is that we may need several reduction functions, but seems that summary only does means of values so it should be generic enough and we can implement it without modifying summary?

This is just a non-well thought idea so feel free to ignore it if it doesn't make sense :)

emcastillo (commented):
This is great!!!
I just have some minor concerns but I would say this is mostly good to go.

PR #552, opened 2022-05-26
Fix `protobuf` dependency for onnx

emcastillo:


emcastillo:
Rebuilding image

https://ci.preferred.jp/pytorch-pfn-extras.prep-linux/102783/

emcastillo:
I think this PR is no longer needed

PR #553, opened 2022-06-30
Support exporting testcases with scripted modules

takagi:
Currently, `onnx.export_testcase` seems not to support exporting test cases with scripted (not traced) TorchScript modules, that is, instances of `torch.jit.ScriptedModule`.

I found its `as_output` feature prevented it from exporting scripted modules, and I've fixed it just to bypass `_ModuleWithAdditionalOutputs` as, I suppose, `as_output` feature essentially does not work with scripting at the same time.

### DIFF
--- a/pytorch_pfn_extras/onnx/_as_output.py
+++ b/pytorch_pfn_extras/onnx/_as_output.py
@@ -3,6 +3,7 @@ from typing import Any, Generator, List, NamedTuple, Tuple, Union
 import torch
 import threading
 from contextlib import contextmanager
+import warnings
 
 _outputs = threading.local()
 
@@ -99,6 +100,9 @@ def trace(
 
 def as_output(name: str, value: torch.Tensor) -> torch.Tensor:
     if torch.jit.is_scripting():
+        warnings.warn(
+            '`as_output` seen in TorchScript compilation. The value is no '
+            'longer an output in the exported onnx.')
         return value
     if hasattr(_outputs, "outputs") and _outputs.outputs is not None:
         _outputs.outputs.add(name, value)
--- a/pytorch_pfn_extras/onnx/_as_output.py
+++ b/pytorch_pfn_extras/onnx/_as_output.py
@@ -1,5 +1,5 @@
 import onnx
-from typing import Any, Generator, List, NamedTuple, Tuple
+from typing import Any, Generator, List, NamedTuple, Tuple, Union
 import torch
 import threading
 from contextlib import contextmanager
@@ -85,10 +85,11 @@ class _ModuleWithAdditionalOutputs(torch.nn.Module):
 
 @contextmanager
 def trace(
-        module: torch.nn.Module
+        module: Union[torch.nn.Module, torch.jit.ScriptModule]
 ) -> Generator[Tuple[torch.nn.Module, _Outputs], None, None]:
     _outputs.outputs = _Outputs()
-    module = _ModuleWithAdditionalOutputs(module, _outputs.outputs)
+    if not isinstance(module, torch.jit.ScriptModule):
+        module = _ModuleWithAdditionalOutputs(module, _outputs.outputs)
     try:
         yield module, _outputs.outputs
     finally:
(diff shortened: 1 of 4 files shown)


tests/pytorch_pfn_extras_tests/onnx_tests/test_as_output.py
```diff
@@ -95,3 +95,33 @@ def forward(self, x):
...
+            if b:  # IF stmt. to check scripted (not traced) IR
+                h = -h
+            h = as_output("h", h)
+            h = self.linear(h)
+            return h
+
+    model = torch.jit.script(Net())
+    x = torch.ones((1, 1, 32, 32))
+    b = torch.tensor(True)
+    output_dir = _helper(model, (x, b), 'as_output', use_pfto=False)
```
takagi, line 119: Seems pfto_exporter dose not support scripting as it calls `torch.jit.trace`. Maybe we should also fix it, but I'm not confident about it as I'm not sure the purpose of pfto_exporter.

pytorch_pfn_extras/onnx/_as_output.py
```diff
@@ -97,6 +98,8 @@ def trace(
 
 
 def as_output(name: str, value: torch.Tensor) -> torch.Tensor:
+    if torch.jit.is_scripting():
+        return value
```
xuzijian629, line 102: Is it possible to reach here? (I mean, can `as_output` be called during exporing script module?)

pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -283,6 +283,8 @@ def export_testcase(
     onnx_graph, outs = _export(
         model, args, strip_large_tensor_data, large_tensor_threshold,
         input_names=input_names, **kwargs)
+    if isinstance(model, torch.jit.ScriptModule):
+        outs = model(*args)
```
xuzijian629, line 287: https://github.com/pytorch/pytorch/blob/v1.12.0/torch/onnx/utils.py#L717-L718

according to above, the returned `outs` in torch export is None when model is `torch.jit.ScriptModule`.
To clarify that, I think it's better to assert `outs` is None here.

```suggestion
        assert outs is None
        outs = model(*args)
```

pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -306,14 +308,29 @@ def export_testcase(
...
+                        and is_large_tensor(t, large_tensor_threshold)):
+                    _strip_raw_data(t)
+                fp.write(t.SerializeToString())
+        elif isinstance(tensor, Sequence):
+            array = [t.detach().cpu().numpy() for t in tensor]
+            with open(f, 'wb') as fp:
+                s = onnx.numpy_helper.from_list(array, name)
+                if strip_large_tensor_data:
+                    for t in s.tensor_values:
+                        if (is_large_tensor(t, large_tensor_threshold)):
```
xuzijian629, line 329:
```suggestion
                        if is_large_tensor(t, large_tensor_threshold):
```

pytorch_pfn_extras/onnx/_as_output.py
takagi: Calling `as_output` itself is valid for TorchScript. It's in case, I suppose, a user has a module that contains `as_output` calls, and he/she could export the module without removing the `as_output` calls every time. But, yeah, we can remove it if it is uncommon use.

xuzijian629: I see. Then, how about show a warning in as_output if called from script module saying that the value is no longer an output in the exported onnx?

takagi: Fine, I'll make it as such!

takagi:
PTAL!

xuzijian629 (approved):
LGTM!! Thanks for this great feature!

takagi:
I'm seeing a style check failure.

xuzijian629 (approved):

PR #558, opened 2022-07-07
Separate torch onnx constants for version compatibility

xuzijian629:
From torch 1.12, many constants and global variables were moved from symbolic_helper to `torch.onnx._constants` and `torch.onnx._globals`. I created `pytorch_pfn_extras.onnx.globals` to absorb these differences.


xuzijian629:
`GLOBALS` members are not constant, so we have to fetch the latest value before return. For example, `export_opset_version` can be updated by calling 
https://github.com/pfnet/pytorch-pfn-extras/blob/b0fb3e94bbbefdac86e3ab73f22ba830c11e980a/pytorch_pfn_extras/onnx/pfto_exporter/export.py#L865 

`CONSTANTS` members constant.

xuzijian629:
Maybe we should also implement setters to GLOBALS

pytorch_pfn_extras/onnx/globals.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @property
+    def onnx_shape_inference(self) -> bool:
+        if pytorch_pfn_extras.requires("1.12.0"):
+            return torch.onnx._globals.onnx_shape_inference  # type: ignore[no-any-return]
+        else:
+            return torch.onnx.symbolic_helper._onnx_shape_inference
+
+
+GLOBALS = _TorchOnnxGlobals()
```
asi1024, line 53: How about simplify as follows?
```suggestion
if pytorch_pfn_extras.requires("1.12.0"):
    GLOBALS = torch.onnx._globals
else:
    GLOBALS = torch.onnx.symbolic_helper
```

asi1024: `GLOBALS = torch.onnx._globals.GLOBALS` is correct?

xuzijian629: Thanks!

For torch 1.11 and before, the global variables begin with `_` like `torch.onnx.symbolic_helper._export_onnx_opset_version`, and different from torch 1.12's `GLOBAL.export_onnx_opset_version`. So I created a class `_InternalGlobalsBeforeTorch1_11`.

pytorch_pfn_extras/onnx/globals.py
```diff
@@ -0,0 +1,53 @@
...
+            self.default_opset_version = torch.onnx.symbolic_helper._default_onnx_opset_version
+            self.main_opset_version = torch.onnx.symbolic_helper._onnx_main_opset
+            self.stable_opset_versions = torch.onnx.symbolic_helper._onnx_stable_opsets
+            if pytorch_pfn_extras.requires("1.11.0"):
+                self.constant_folding_opset_versions = torch.onnx.symbolic_helper._constant_folding_opset_versions  # type: ignore[attr-defined]
+            else:
+                self.constant_folding_opset_versions = torch.onnx.constant_folding_opset_versions  # type: ignore[attr-defined]
+
+
+CONSTANTS = _TorchOnnxConstants()
```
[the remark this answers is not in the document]
xuzijian629: To follow torch 1.12's style, I created `pytorch_pfn_extras.onnx._constants` like `torch.onnx._constants` and put constants there.

asi1024 (approved):

PR #559, opened 2022-07-07
[PFTO] Save denotation with uppercase

xuzijian629:
ONNX standard denotations are uppercase.
https://github.com/onnx/onnx/blob/02d8f9626f1352e3837d35f45f2f37270e4b85a1/docs/TypeDenotation.md#type-denotation-definition

Especially, when visualizing ONNX with netron, denotation string is case-sensitive.
https://github.com/lutzroeder/netron/blob/4fe5bfc790be57c6b107877b66d583d8192d7b8d/source/onnx.js#L1656


take-cheeze (approved):
👍

PR #560, opened 2022-07-07
[PFTO] Set domain for custom onnx nodes

xuzijian629:
# Problem
With current master's PFTO exporter, exporting a node with `org.Chainer::RandInt` produces a onnx node without `domain` attribute.
In such case, onnx checker raises error because the exported node will be considered to have default ("onnx") domain and there is no RandInt operator in onnx domain.

We need to set appropriate domain for custom operators.

### DIFF
--- a/pytorch_pfn_extras/onnx/pfto_exporter/export.py
+++ b/pytorch_pfn_extras/onnx/pfto_exporter/export.py
@@ -850,6 +850,7 @@ class _Exporter(_ExporterOptions):
         model: onnx.ModelProto = onnx.helper.make_model(
             graph,
             opset_imports=[onnx.helper.make_opsetid("", self.opset_version)],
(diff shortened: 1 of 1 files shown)


pytorch_pfn_extras/onnx/pfto_exporter/export.py
```diff
@@ -847,6 +850,7 @@ def apply_dynamic_axes_info(out: onnx.ValueInfoProto, k: str) -> None:
         model: onnx.ModelProto = onnx.helper.make_model(
             graph,
             opset_imports=[onnx.helper.make_opsetid("", self.opset_version)],
+            producer_name="pfto",
```
take-cheeze, line 853: This seems to be unrelated but sorry for forgetting this... 👍

asi1024 (approved):

PR #566, opened 2022-07-08
Update import locations for torch 1.12

xuzijian629:
https://github.com/pfnet/pytorch-pfn-extras/pull/558 creates a class for constant and global values to absorb the difference between torch 1.11 and 1.12, but we may need to think carefully whether such a class is really important or not.

As a simple alternative, we can just change import locations based on torch version.

If this PR will be merged, then close #558


xuzijian629:
we discussed offline and decided to continue #558

PR #584, opened 2022-08-04
Add interface to trace events in custom runtimes

emcastillo:


pytorch_pfn_extras/runtime/_runtime.py
```diff
@@ -311,6 +311,17 @@ def map(
         """
         raise NotImplementedError()
 
+    @classmethod
+    @contextlib.contextmanager
+    def trace(cls, event_name: str, arg: Any) -> Generator[None, None, None]:
```
emcastillo, line 316: This is a class method since we can have multiple runtime instances (one per module created with `ppe.to`). I think we should address these issues in other PR since the ppe.to and nested modules semantics changed and can be tricky.

pytorch_pfn_extras/training/_trainer.py
```diff
@@ -218,6 +218,8 @@ def run(self,
         self._train_len = train_len
         self._eval_len = eval_len
 
+        device = self.handler._entry_runtime.device_spec  # type: ignore[attr-defined]
```
emcastillo, line 221: wonder if we should pass the device spec as a trainer arg instead of accessing it like this. It would be much simpler.

kmaehashi:
Sorry for the delay, LGTM!
/test

kmaehashi (approved):

PR #586, opened 2022-08-10
Add Slack app manifest

kmaehashi:
Add an example app manifest to allow users setup the app quickly.


emcastillo (approved):
LGTM!
I thought that this needed to be manually set up via the web interface.

PR #587, opened 2022-08-10
Fix docs for Slack extension

kmaehashi:


emcastillo (approved):
Checked rendered docs!

PR #597, opened 2022-08-25
Fix import failure when user account does not exist

kmaehashi:
```
>>> import getpass
>>> getpass.getuser()
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.10/getpass.py", line 169, in getuser
    return pwd.getpwuid(os.getuid())[0]
KeyError: 'getpwuid(): uid not found: 12345'
```


toslunar (approved):
LGTM

emcastillo (approved):

PR #497, opened 2022-01-21
Support torch 1.10 on onnx export

emcastillo:
still wip, need to adjust several parameters depending on torch version

Fixes #492

PR #602, opened 2022-10-04
[onnx] Disable logging in non-verbose

take-cheeze:
From torch 1.10 support we are passing `verbose=True` to enable doc_string but it makes stdout dirty so here is a workaround for it using torch.onnx logging feature introduced in torch 1.12.

- Ref https://github.com/pytorch/pytorch/commit/54a6942f8ddc5e9d5277938e163453e110432c5a
- Ref https://github.com/pfnet/pytorch-pfn-extras/pull/497

### DIFF
--- a/pytorch_pfn_extras/onnx/export_testcase.py
+++ b/pytorch_pfn_extras/onnx/export_testcase.py
@@ -157,7 +157,8 @@ def _export(
     bytesio = io.BytesIO()
     opset_ver = kwargs.get('opset_version', None)
     force_verbose = False
-    original_log = torch.onnx.log
+    if pytorch_pfn_extras.requires('1.12.0'):
+        original_log = torch.onnx.log
     if opset_ver is None:
         opset_ver = pytorch_pfn_extras.onnx._constants.onnx_default_opset
         kwargs['opset_version'] = opset_ver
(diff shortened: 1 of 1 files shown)


pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -164,6 +166,15 @@ def _export(
         kwargs['strip_doc_string'] = False
     else:
         strip_doc_string = kwargs.pop('strip_doc_string', True)
+        if not kwargs.get('verbose', False):
+            force_verbose = True
+            if pytorch_pfn_extras.requires('1.12.0'):
+                #  Following line won't work because verbose mode always
+                # enable logging so we are replacing python function instead:
+                # torch.onnx.disable_log()
```
take-cheeze, line 174: Ref: https://github.com/pytorch/pytorch/blob/b95e0fcc2c40f08f43cc69edbb0168ab17facbda/torch/onnx/utils.py#L167-L168

xuzijian629 (dismissed):
LGTM!

pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -185,6 +198,10 @@ def _export(
     if strip_large_tensor_data:
         _strip_large_initializer_raw_data(onnx_graph, large_tensor_threshold)
 
+    if force_verbose:
+        # torch.onnx.enable_log()
+        torch.onnx.log = original_log  # type: ignore[attr-defined]
```
xuzijian629, line 203:
```suggestion
    if force_verbose:
        if pytorch_pfn_extras.requires('1.12.0'):
            # torch.onnx.enable_log()
            torch.onnx.log = original_log  # type: ignore[attr-defined]
```

xuzijian629 (approved):

PR #603, opened 2022-10-06
[torchscript] Add torchscript utility to find inplace op

take-cheeze:
Inplace binary operator is sometimes hard find from but it could be easily found after compiling to torchscript


kmaehashi (approved):
LGTM!

PR #604, opened 2022-10-10
Always create expected output for export_testcase

shinh:
`_export_util` should always return outputs for `export_testcase`. Re-run the model to create outputs when ONNX checker fails.

### DIFF
--- a/pytorch_pfn_extras/onnx/export_testcase.py
+++ b/pytorch_pfn_extras/onnx/export_testcase.py
@@ -155,7 +155,7 @@ def _export(
         strip_large_tensor_data: bool = False,
         large_tensor_threshold: int = LARGE_TENSOR_DATA_THRESHOLD,
         use_pfto: bool = False,
-        return_output: bool = False,
+        return_output: bool = True,
         **kwargs: Any,
 ) -> Tuple[onnx.ModelProto, Any]:
     model.zero_grad()
--- a/pytorch_pfn_extras/onnx/export_testcase.py
+++ b/pytorch_pfn_extras/onnx/export_testcase.py
@@ -94,6 +94,7 @@ def _export_util(
         model: torch.nn.Module,
         args: Sequence[Any],
         f: IO,
+        return_output: bool = False,
         **kwargs: Any,
 ) -> Any:
     """Wrap operator type to export
@@ -137,6 +138,9 @@ def _export_util(
             except checker_error:
                 if enable_onnx_checker:
                     raise
+                if return_output:
+                    # Re-run the model to obtain the output.
+                    return model(*args)
         else:
             kwargs['_retain_param_name'] = True
             return torch_export(  # type: ignore[no-untyped-call]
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -137,6 +138,9 @@ def _export_util(
             except checker_error:
                 if enable_onnx_checker:
                     raise
+                if return_output:
+                    # Re-run the model to obtain the output.
+                    return model(*args)
```
xuzijian629, line 143: Do you still need the outputs even when the export fails?

shinh: Thanks for the question! Historically, `torch.onnx.export` had a kwarg `enable_onnx_checker` which turns off ONNX checker. As the kwarg is removed, we are emulating the behavior here by catching `torch.onnx.utils.ONNXCheckerError`. The emulation was working good for `ppe.onnx.export` as it does not need outputs but was broken for `ppe.onnx.export_testcase`. Without this PR, `test_data_set_0/output_0.pb` will be missing if we export ONNX-incompliant ONNX.

pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -151,6 +155,7 @@ def _export(
         strip_large_tensor_data: bool = False,
         large_tensor_threshold: int = LARGE_TENSOR_DATA_THRESHOLD,
         use_pfto: bool = False,
+        return_output: bool = False,
```
take-cheeze, line 158: This should be true for compatibility with torch.onnx because they always returns `torch_out` except `torch.onnx.export`
https://sourcegraph.com/github.com/pytorch/pytorch@4bfb7341819b3bfcaf65ddc136f25d23983740a7/-/blob/torch/onnx/utils.py?L1663
https://sourcegraph.com/github.com/pytorch/pytorch@4bfb7341819b3bfcaf65ddc136f25d23983740a7/-/blob/torch/onnx/utils.py?L208

shinh: Thanks, done!

pytorch_pfn_extras/onnx/export_testcase.py
shinh: Also please note `torch.onnx` ensures model export fails due to `CheckerError`:
```
      CheckerError: If the ONNX checker detects an invalid ONNX graph. Will still export the model to the file ``f`` even if this is raised.
```

take-cheeze (approved):
LGTM 👍

xuzijian629 (dismissed):
LGTM!

shinh:
`ppe.onnx.grad` was failing due to domain mismatch: https://github.com/pfnet/pytorch-pfn-extras/pull/604/commits/a593b7f2ceffa61b7ec0833f0a61b3a7e2425451

xuzijian629 (approved):

PR #614, opened 2022-10-26
[CI] Limit branch to run GitHub Actions

take-cheeze:
Sometimes running on experimental commit push is noisy


kmaehashi:
From the offline discussion, closing this PR as the current configuration can be benefitial for contributors to run tests on their own branch in forks. Let's reconsider when actions runners starts to clog.

PR #615, opened 2022-10-26
[CI] Test with nightly torch daily

take-cheeze:
By using nightly built torch wheel
With `workflow_dispatch` this job could be dispatched manually from actions tab. Ref: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow


kmaehashi (approved):
Looks good!

PR #621, opened 2022-10-31
[WIP] Disable ONNX for torch 1.13

emcastillo:


take-cheeze (approved):
LGTM 👍 
Confirmed that `python3 -c "import pytorch_pfn_extras.onnx"` works fine in torch 1.13

pytorch_pfn_extras/onnx/pfto_exporter/export.py
```diff
@@ -14,11 +14,18 @@
 from pytorch_pfn_extras.torchscript import run_jit_pass
 import torch
 import torch.jit
-import torch.onnx.symbolic_helper as sym_hel
```
take-cheeze, line 17: Just some additional note that `symbolic_helper` module is still alive: https://github.com/pytorch/pytorch/blob/release/1.13/torch/onnx/symbolic_helper.py
Though it doesn't change any test result so will ignore for now

PR #631, opened 2022-11-18
[onnx] Avoid circular import by lazy import

take-cheeze:

PR #633, opened 2022-11-21
[pfto] Lazy import symbolic registry for circular dependency

take-cheeze:
Rel https://github.com/pfnet/pytorch-pfn-extras/pull/631


kmaehashi (approved):
LGTM!

PR #657, opened 2023-04-03
Add a Logic class for clousure mode optimization.

### LINKED ISSUE #636: Support passing `closure` argument to optimizers
Optimizers supports two kind of `step()` call:

https://pytorch.org/docs/stable/optim.html#taking-an-optimization-step

However, some optimizers require mandatory `closure` argument:

https://pytorch.org/docs/stable/generated/torch.optim.LBFGS.html#torch.optim.LBFGS

Currently, ExtensionsManager and Logic do not support these optimizers.

linshoK:
closes #636


pytorch_pfn_extras/handler/_logic.py
```diff
@@ -464,3 +465,198 @@ def eval_step(
...
+@dataclasses.dataclass
+class ClousureModelOutput:
+    outs: Any
+    loss: torch.Tensor
+
+    def __float__(self) -> float:
+        return float(self.loss)
+
+
+class ClousureLogic(BaseLogic):
```
emcastillo, line 479:
```suggestion
class ClousureLogic(Logic):
```

Maybe it is better just to specialize `Logic` so we can avoid repeating all the methods :)
I think with this, only `train_step` becomes necessary.

linshokaku: Corrected the relevant code.

pytorch_pfn_extras/handler/_logic.py
```diff
@@ -464,3 +465,80 @@ def eval_step(
...
+    loss: torch.Tensor
+
+    def __float__(self) -> float:
+        return float(self.loss)
+
+
+class ClousureLogic(Logic):
+
+    def consume_options(self, options: Dict[str, Any]) -> None:
+        super().consume_options(options)
```
emcastillo, line 482: Can this function be deleted too? :)

emcastillo, line 482: Can this function be deleted too? :)

emcastillo, line 482: Can this function be deleted too? :)

linshokaku: I can't remove this method completely because it changes the process when grad_scaler options are passed.

Most of it is the same as Logic.cosnsume_options, so calling it from the super class is almost sufficient, so most of it can be removed. We will modify this one.

linshokaku: https://github.com/pfnet/pytorch-pfn-extras/issues/636#issuecomment-1493696719

We have confirmed that grad_scaler and clousure step cannot be used at the same time.

emcastillo (approved):

pytorch_pfn_extras/handler/_logic.py
```diff
@@ -464,3 +465,80 @@ def eval_step(
...
+
+        If you want to change the stepping of the optimizer, please use the normal Logic class.
+        
+        Args:
+            optimizers (dict of torch.optim.Optimizer):
+                The optimizers.
+            batch_idx (int):
+                Number of steps already finished.
+        """
+        pass
```
linshokaku, line 544: The train_step_optimizers method of the parent class should not be executed, so the method is inherited and the execution is empty.

PR #666, opened 2023-04-13
[onnx] Parameterize onnx grad test with pfto

take-cheeze:


hvy:
In my understanding, we can fix the type of `zs` to become `onnx.AttributeProto.STRINGS` using `[""]`. There might be a cleaner solution though 🙇 

Also, it seems like pfto calls forward twice during the ONNX export. 
The second time is not during tracing, and the thread local output container is cleared and names becomes inconsistent.
Is this clearing here necessary? Cc. @HiroakiMikami (see diff below)
By not clearing the container, we get the correct names.
I'd appreciate any discussion since I'm not that familiar with other parts of the code base (maybe this'll break assumptions elsewhere). 
I could send a PR against this PR-branch if it seems reasonable. Thanks.

```diff
diff --git a/pytorch_pfn_extras/onnx/_as_output.py b/pytorch_pfn_extras/onnx/_as_output.py
index a723868..baa40fb 100644
--- a/pytorch_pfn_extras/onnx/_as_output.py
+++ b/pytorch_pfn_extras/onnx/_as_output.py
@@ -60,21 +60,19 @@ class _Outputs:


 class _ModuleWithAdditionalOutputs(torch.nn.Module):
-    def __init__(self, module: torch.nn.Module, outputs: _Outputs) -> None:
+    def __init__(self, module: torch.nn.Module) -> None:
         super().__init__()  # type: ignore[no-untyped-call]
         self._ppe_as_out_module: torch.nn.Module = module
-        self.outputs = outputs

     def forward(self, *args: Any, **kwargs: Any) -> Any:
-        self.outputs.clear()
         out = self._ppe_as_out_module(*args, **kwargs)
-        if len(self.outputs.values) == 0:
+        if len(_outputs.outputs.values) == 0:
             return out
         if isinstance(out, torch.Tensor):
             out = [out]
         elif not isinstance(out, list):
             out = list(out)
-        out.extend([value for _, value in self.outputs.values])
+        out.extend([value for _, value in _outputs.outputs.values])
         return tuple(out)

     def state_dict(self, *args, **kwargs):  # type: ignore[no-untyped-def]
@@ -90,7 +88,7 @@ def trace(
 ) -> Generator[Tuple[torch.nn.Module, _Outputs], None, None]:
     _outputs.outputs = _Outputs()
     if not isinstance(module, torch.jit.ScriptModule):
-        module = _ModuleWithAdditionalOutputs(module, _outputs.outputs)
+        module = _ModuleWithAdditionalOutputs(module)
     try:
         yield module, _outputs.outputs
     finally:
diff --git a/pytorch_pfn_extras/onnx/_grad.py b/pytorch_pfn_extras/onnx/_grad.py
index 244976f..361c265 100644
--- a/pytorch_pfn_extras/onnx/_grad.py
+++ b/pytorch_pfn_extras/onnx/_grad.py
@@ -85,10 +85,10 @@ def grad(
             @staticmethod
             def symbolic(g, output, grad_output, *inputs):  # type: ignore
                 return g.op(
-                    "ai.onnx.preview::Gradient",
+                    "ai.onnx.preview.training::Gradient",
                     *inputs,
                     xs_s=input_names,
-                    zs_s=[],
+                    zs_s=[""],
                     y_s=output_name,
                     outputs=len(input_names),
                 )
```

tests/pytorch_pfn_extras_tests/onnx_tests/test_grad.py
```diff
@@ -110,21 +111,34 @@ def forward(self, x):
...
+        assert list([v.name for v in actual_onnx.graph.output]) == [
+            "linear.72", "Gradient_y_0", "Gradient_x_0_0"
+        ]
+        y_in, _ = _get_name(actual_onnx.graph, "input.1")
+        if pytorch_pfn_extras.requires("1.13"):
+            assert named_nodes["/_ppe_as_out_module/conv/Conv"].input[0] == "Gradient_x_0_0"
+            assert named_nodes["/_ppe_as_out_module/conv/Conv"].output[0] == y_in
+        else:
+            assert named_nodes["Conv_2"].input[0] == "Gradient_x_0_0"
+            assert named_nodes["Conv_2"].output[0] == y_in
```
hvy, line 124: How about also checking the attributes, at least of `Gradient_4`?

take-cheeze:
@hvy
Thank you for investegation

Since `_ModuleWithAdditionalOutputs` runs
```python
self.outputs.clear()
```
in `forward` method getting original outputs before tracing will get the correct output infos so https://github.com/pfnet/pytorch-pfn-extras/pull/666/commits/771b38ba1f33e1ccc8a84b3c1809ca44dd996519 should fix the issue.

emcastillo (approved):

PR #672, opened 2023-04-21
[pfto] Run CSE to reduce unnecessary constants

take-cheeze:
Ref https://github.com/pytorch/pytorch/pull/85665


xuzijian629 (approved):
LGTM!

PR #678, opened 2023-05-02
[pfto] Force is_tracing() == True in test outputs generation

take-cheeze:


pytorch_pfn_extras/onnx/pfto_exporter/export.py
```diff
@@ -237,7 +252,8 @@ def _run_trace(self) -> None:
 
         # TODO(twata): Use `self.traced` instead or use traced result outputs
         self._restore_state()
-        self.original_outputs = self.original_model(*self.inputs)
+        with _force_tracing():
+            self.original_outputs = self.original_model(*self.inputs)
```
xuzijian629, line 256: memo: this context manager is designed to guarantee that the collected outputs actually correspond to onnx outputs

xuzijian629 (approved):
LGTM!

PR #681, opened 2023-05-09
Add cifar10 example

linshokaku:
Training code using CIFAR10 will be added, including both regular training on a single GPU and training using distributed data parallel with a trainer.


emcastillo:
Can we try to run it in the CI? or does it take too long :)

https://github.com/pfnet/pytorch-pfn-extras/blob/master/.flexci/linux/script.sh#L64

linshokaku:
unittest Added!

PR #691, opened 2023-05-18
refactor gradscaler assertion

linshokaku:


kmaehashi:
This allows autocast to work when `len(backward_outputs) == 1`.

PR #692, opened 2023-05-18
Also look for snapshot in ``writer.out_dir``

puchupala:
We save our snapshots in `snapshot` sub-directory under `outdir`. When we set our snapshot directory to `writer.out_dir`, the snapshots are saved properly but never got loaded.

### DIFF
--- a/pytorch_pfn_extras/training/extensions/_snapshot.py
+++ b/pytorch_pfn_extras/training/extensions/_snapshot.py
@@ -337,7 +337,6 @@ class _Snapshot(extension.Extension):
     def initialize(  # type: ignore[override]
             self, manager: ExtensionsManagerProtocol) -> Optional[str]:
         target = manager if self._target is None else self._target
-        outdir = manager.out
         writer = manager.writer if self.writer is None else self.writer
         self.writer = writer
         loaded_fn = None
@@ -376,10 +375,10 @@ class _Snapshot(extension.Extension):
             # injected here.
             def _cleanup() -> None:
                 assert writer is not None
-                files = _find_stale_snapshots(self.filename, outdir,
+                files = _find_stale_snapshots(self.filename, writer.out_dir,
                                               self.n_retains, writer.fs)
                 for file in files:
-                    writer.fs.remove(os.path.join(outdir, file))
+                    writer.fs.remove(os.path.join(writer.out_dir, file))
 
             assert writer is not None
             writer._add_cleanup_hook(_cleanup)
--- a/pytorch_pfn_extras/training/extensions/_snapshot.py
+++ b/pytorch_pfn_extras/training/extensions/_snapshot.py
@@ -342,7 +342,7 @@ class _Snapshot(extension.Extension):
         self.writer = writer
         loaded_fn = None
         if self.autoload:
-            # If ``autoload`` is on, this code scans the ``outdir``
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/training/extensions/_snapshot.py
```diff
@@ -349,6 +349,9 @@ def initialize(  # type: ignore[override]
             # manager.
             assert writer is not None
             loaded_fn = _find_latest_snapshot(self.filename, outdir, writer.fs)
+            if not loaded_fn:
+                # Try searching in ``writer.out_dir`` if there is no snapshot in ``outdir``
+                loaded_fn = _find_latest_snapshot(self.filename, writer.out_dir, writer.fs)
```
linshokaku, line 354:
```suggestion
            loaded_fn = _find_latest_snapshot(self.filename, writer.out_dir, writer.fs)
            if not loaded_fn:
                loaded_fn = _find_latest_snapshot(self.filename, outdir, writer.fs)
```

When a snapshot object has a writer object, I want the value of the out_dir property of the writer object to be used first, so I want the order to be switched.
When the snapshot object does not have a writer object, the value of `outdir` matches `manager.writer.out_dir`, so I think it is ok to change as suggested.

linshokaku (changes requested):

pytorch_pfn_extras/training/extensions/_snapshot.py
```diff
@@ -348,9 +348,9 @@ def initialize(  # type: ignore[override]
             # terms of mtime, and tries to load it it the target or
             # manager.
             assert writer is not None
-            loaded_fn = _find_latest_snapshot(self.filename, outdir, writer.fs)
+            loaded_fn = _find_latest_snapshot(self.filename, writer.out_dir, writer.fs)
```
puchupala, line 351: Are we no longer searching in `outdir`? (`manager.out`). Also, we might want to update the comment above this line accordingly.

tests/pytorch_pfn_extras_tests/training_tests/extensions_tests/test_snapshot.py
```diff
@@ -384,3 +390,38 @@ def get_epoch_indices():
...
+
+
+def test_snapshot_autoload_initialize_without_writer(path):
+    snapshot_filename = "snapshot_file"
+    trainer = get_trainer(out_dir=path)
+    trainer.models["main"]._state_dict = {"value": 0}
+
+    snapshot = extensions.snapshot(filename=snapshot_filename)
+    snapshot(trainer)
+    assert os.path.isfile(f"{path}/{snapshot_filename}")
```
puchupala, line 402: Concat path with `/` string is rather unsafe. You might want to use `os.path.join()` instead?

linshokaku: I agree. I fixed the code.

pytorch_pfn_extras/training/extensions/_snapshot.py
linshokaku: Fixed it.

linshokaku:
@puchupala 
Now that I have made the necessary corrections and added the tests, may I review the changes and get your comments?

puchupala (commented):
LGTM! Thank you very much for the tests!

PR #706, opened 2023-06-05
Drop torch 1.9 from CIs and add Python 3.8

emcastillo:


kmaehashi:
Removed webhook and branch protection settings for PyTorch 1.9 CIs.

Issue #717, opened 2023-06-28
Make `export_testcase` API's `args` behave the same as `torch.onnx.export`


kmaehashi:
`args` supports Tensor, tuple of Tensors, or tuple like `(args..., kwargs)`.

https://pytorch.org/docs/stable/onnx.html#functions

PR #718, opened 2023-06-30
Support kwargs in export_testcase

xuzijian629:
close: #717 

supported `export_testcase` with kwargs.
Similarly to https://pytorch.org/docs/stable/onnx.html#torch.onnx.export , you can do this by putting a dictionary of kwargs at the last element of args.

When doing this, users are required to explicitly specify `input_names` in order to name input pb tensors (kwargs are not positional, so we don't use input_0.pb-like naming).

I also added these remarks in the docs.

### DIFF
--- a/pytorch_pfn_extras/onnx/export_testcase.py
+++ b/pytorch_pfn_extras/onnx/export_testcase.py
@@ -16,7 +16,7 @@ import torch
 import torch.autograd
 from torch.onnx import OperatorExportTypes
 from torch.onnx.utils import \
-    _export as torch_export, _model_to_graph as torch_model_to_graph
+    _export as torch_export, _model_to_graph as torch_model_to_graph, _decide_input_format
 
 from pytorch_pfn_extras.onnx import _as_output as as_output
 from pytorch_pfn_extras.onnx import _grad as grad
@@ -191,6 +191,8 @@ def _export(
 
                 torch.onnx.log = no_op  # type: ignore[attr-defined]
         kwargs['verbose'] = True
+    # Exted args with kwargs (including default values)
+    args = _decide_input_format(model, args)  # type: ignore[no-untyped-call]
     with init_annotate(model, opset_ver) as ann, \
             as_output.trace(model) as (model, outputs), \
             grad.init_grad_state(), \
@@ -305,6 +307,13 @@ def export_testcase(
 
     .. warning:: This function is not thread safe.
 
+    .. note:: 
+        When exporting a model whose forward takes keyword arguments of ``torch.Tensor`` type,
+        you can pass them by putting a dict as the last element of ``args``.
+        When the keyword arguments have default values, you need to explicitly include
+        them into the dict.
+        Also, you must explicitly specify ``input_names`` that are the names of both positional
+        and keyword arguments.
     """
 
     if user_meta is None:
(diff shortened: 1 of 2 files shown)


tests/pytorch_pfn_extras_tests/onnx_tests/test_export_testcase.py
```diff
@@ -494,3 +494,117 @@ def forward(self, x, b):
...
+
+    # Test with labels
+    output_dir = _get_output_dir('export_kwargs')
+    export_testcase(
+        model,
+        (x, {"labels": l}),
+        output_dir,
+        input_names=["x", "labels"],
+        training=model.training,
+        do_constant_folding=False,
```
take-cheeze, line 541:
```suggestion
        do_constant_folding=False,
        opset_version=12,
```

tests/pytorch_pfn_extras_tests/onnx_tests/test_export_testcase.py
```diff
@@ -494,3 +494,117 @@ def forward(self, x, b):
...
+
+    # Test without labels
+    output_dir = _get_output_dir('export_kwargs2')
+    export_testcase(
+        model,
+        (x,),
+        output_dir,
+        input_names=["x"],
+        training=model.training,
+        do_constant_folding=False,
```
take-cheeze, line 554:
```suggestion
        do_constant_folding=False,
        opset_version=12,
```

tests/pytorch_pfn_extras_tests/onnx_tests/test_export_testcase.py
```diff
@@ -494,3 +494,117 @@ def forward(self, x, b):
...
+
+    # Test with labels
+    output_dir = _get_output_dir('export_default_kwargs')
+    export_testcase(
+        model,
+        (x, {"labels": l, "bias": torch.zeros(10)}),
+        output_dir,
+        input_names=["x", "labels", "bias"],
+        training=model.training,
+        do_constant_folding=False,
```
take-cheeze, line 594:
```suggestion
        do_constant_folding=False,
        opset_version=12,
```

tests/pytorch_pfn_extras_tests/onnx_tests/test_export_testcase.py
```diff
@@ -494,3 +494,117 @@ def forward(self, x, b):
...
+
+    # Test without labels
+    output_dir = _get_output_dir('export_default_kwargs2')
+    export_testcase(
+        model,
+        (x, {"bias": torch.ones(10)}),
+        output_dir,
+        input_names=["x", "bias"],
+        training=model.training,
+        do_constant_folding=False,
```
take-cheeze, line 607:
```suggestion
        do_constant_folding=False,
        opset_version=12,
```

pytorch_pfn_extras/onnx/export_testcase.py
```diff
@@ -313,10 +322,24 @@ def export_testcase(
...
+        
+    def has_kwargs_in_args(args: tuple) -> bool:
+        return len(args) >= 1 and isinstance(args[-1], dict)
+
+    if has_kwargs_in_args(args):
+        assert "input_names" in kwargs, 'export_testcase needs explicit "input_names" when exporting with kwargs'
+        named_args_list = list(args[:-1])
+        for key, tensor in args[-1].items():
+            assert isinstance(tensor, torch.Tensor)
+            named_args_list.append(tensor)
```
take-cheeze, line 334: I feel implicit conversion seems to be useful for python scalar inputs
```suggestion
            if not isinstance(tensor, torch.Tensor):
                tensor = torch.tensor(tensor)
            named_args_list.append(tensor)
```

take-cheeze (approved):
LGTM 👍 
I've fixed opset version issue in old torch

pytorch_pfn_extras/onnx/export_testcase.py
xuzijian629: I don't fully understand current behavior of export_testcase when a python scalar is given.
In export with testcase, we need to ensure that the number of onnx inputs and the number of testcase's input pb tensors match. If we implicitly create a tensor, doesn't it break this constraint?

I think it'll be safer if we just accept explicit tensor in such cases.

xuzijian629:
Thanks a lot to fix the bug of opset version!

pytorch_pfn_extras/onnx/export_testcase.py
take-cheeze: OK. Your decision seems to be more reasonable! Thank you!

PR #735, opened 2023-07-31
Remove no use out_dir

linshokaku:
Remove the out_dir argument of Writer.__call__ since it does not affect the behavior of the program.


emcastillo:
Discussed offline:

1- This PR is non-compat as some users may directly call `writer.save` in Extensions or related code
2- In order to avoid disruption there are some alternatives
   - Keep the original interface but add a deprecation warning
   - Make the `out_dir` parameter optional (`None`) and use it to override the `Writer.out_dir` attribute when needed

PR #737, opened 2023-08-09
Add wait_for_first_optimizer_step mode

linshokaku:
We will add a mode to delay the execution of scheduler.step until optimizer.step is run. This is to prevent a PyTorch warning that can occur when scheduler.step is executed before optimizer.step, a situation that can arise when GradScaler does not execute optimizer.step on the first iteration.

### DIFF
--- a/pytorch_pfn_extras/training/extensions/lr_scheduler.py
+++ b/pytorch_pfn_extras/training/extensions/lr_scheduler.py
@@ -48,6 +48,9 @@ class LRScheduler(extension.Extension):
         stepper (callable): Function that performs the step on
             the scheduler.
         trigger: Frequency to call this extension.
+        wait_for_first_optimizer_step (bool): Wait until optimizer.step is called
+            before invoking scheduler.step. This can address the issue where
+            optimizer.step is not called from the first iteration when using GradScaler.
     """
 
     def __init__(
@@ -56,14 +59,25 @@ class LRScheduler(extension.Extension):
         *,
         stepper: Any = _default_stepper,
         trigger: trigger_module.TriggerLike = (1, "epoch"),
+        wait_for_first_optimizer_step: bool = False,
         is_async: bool = True,
     ) -> None:
         self.scheduler = scheduler
         self.trigger = trigger_module.get_trigger(trigger)
         self.stepper = stepper
+        self.wait_for_first_optimizer_step = wait_for_first_optimizer_step
+        self._stepped = False
         self.is_async = is_async
 
     def __call__(self, manager: ExtensionsManagerProtocol) -> None:
+        if not self._stepped:
+            if (
+                self.wait_for_first_optimizer_step
+                and self.scheduler.optimizer._step_count < 1
+            ):
+                return
+            self._stepped = True
+
         self.stepper(manager, self.scheduler)
 
     @staticmethod
--- a/pytorch_pfn_extras/training/extensions/lr_scheduler.py
+++ b/pytorch_pfn_extras/training/extensions/lr_scheduler.py
@@ -48,9 +48,9 @@ class LRScheduler(extension.Extension):
         stepper (callable): Function that performs the step on
             the scheduler.
         trigger: Frequency to call this extension.
-        wait_for_first_optimizer_step (bool): Wait until optimizer.step is called
-            before invoking scheduler.step. This can address the issue where
-            optimizer.step is not called from the first iteration when using GradScaler.
+        wait_for_first_optimizer_step (bool): Wait until `optimizer.step()` is called
+            before invoking `scheduler.step()`. This can address the issue where
+            `optimizer.step()` is not called from the first iteration when using GradScaler.
     """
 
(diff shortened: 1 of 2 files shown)


pytorch_pfn_extras/training/extensions/lr_scheduler.py
```diff
@@ -56,14 +59,25 @@ def __init__(
...
         stepper: Any = _default_stepper,
         trigger: trigger_module.TriggerLike = (1, "epoch"),
+        wait_for_first_optimizer_step: bool = False,
         is_async: bool = True,
     ) -> None:
         self.scheduler = scheduler
         self.trigger = trigger_module.get_trigger(trigger)
         self.stepper = stepper
+        self.wait_for_first_optimizer_step = wait_for_first_optimizer_step
+        self._stepped = False
```
kmaehashi, line 69: I guess this flag is essentially unnecessary? Overhead of checking `self.scheduler.optimizer._step_count` for each `__call__` invocation sounds negilible to me.

pytorch_pfn_extras/training/extensions/lr_scheduler.py
```diff
@@ -48,6 +48,9 @@ class LRScheduler(extension.Extension):
         stepper (callable): Function that performs the step on
             the scheduler.
         trigger: Frequency to call this extension.
+        wait_for_first_optimizer_step (bool): Wait until optimizer.step is called
+            before invoking scheduler.step. This can address the issue where
+            optimizer.step is not called from the first iteration when using GradScaler.
```
kmaehashi, line 53:
```suggestion
        wait_for_first_optimizer_step (bool): Wait until `optimizer.step()` is called
            before invoking `scheduler.step()`. This can address the issue where
            `optimizer.step()` is not called from the first iteration when using GradScaler.
```

pytorch_pfn_extras/training/extensions/lr_scheduler.py
```diff
@@ -56,14 +59,25 @@ def __init__(
...
     ) -> None:
         self.scheduler = scheduler
         self.trigger = trigger_module.get_trigger(trigger)
         self.stepper = stepper
+        self.wait_for_first_optimizer_step = wait_for_first_optimizer_step
+        self._stepped = False
         self.is_async = is_async
 
     def __call__(self, manager: ExtensionsManagerProtocol) -> None:
+        if not self._stepped:
```
kmaehashi, line 73: Let's add a comment to say that we rely on the technique as internally used in PyTorch:

```
# https://github.com/pytorch/pytorch/blob/v2.0.1/torch/optim/lr_scheduler.py#L137-L138
```

kmaehashi (changes requested):
LGTM except for nits!

kmaehashi:
For record: `optimizer._step_count` is a private API, but I take it as relatively stable as it's there since the oldest PyTorch version we support (v1.10.0):
https://github.com/pytorch/pytorch/blob/v1.10.0/torch/optim/lr_scheduler.py#L127-L128

pytorch_pfn_extras/training/extensions/lr_scheduler.py
```diff
@@ -71,12 +71,12 @@ def __init__(
     def __call__(self, manager: ExtensionsManagerProtocol) -> None:
         # https://github.com/pytorch/pytorch/blob/v2.0.1/torch/optim/lr_scheduler.py#L137-L138
         if (
-            self.scheduler.optimizer._step_count < 1
-            and self.wait_for_first_optimizer_step
+            self.wait_for_first_optimizer_step
+            and hasattr(self.scheduler.optimizer.step, "_with_counter")
```
kmaehashi, line 75: A very nit: I think assuming the existence of private APIs is fragile, because this feature will silently stop working if PyTorch renames this attribute. I guess this check is not essentially needed and can be removed?

kmaehashi (commented):
Thanks! Could you add tests as well?

pytorch_pfn_extras/training/extensions/lr_scheduler.py
linshokaku: When using the base class of PyTorch's LRScheduler, the addition of the `_step_count` attribute and the addition of a hook for counting are performed at the time of initialization. However, as `ReduceLROnPlateau` does not inherit from this base class, access to these attributes results in an error. Therefore, it is appropriate to perform an existence check for the `_with_counter` attribute, following the code that checks for PyTorch's `lr_scheduler` `_step_count`.

kmaehashi: Ah got it, thanks! I'd prefer this to be commented somewhere, but it can be done separately 😄

linshokaku: I think this may be a very detailed matter and not essential to the functionality of the code, would it be possible to include the link to the comment made on this PR as a comment in the code?

PR #738, opened 2023-08-10
Fix float epoch trigger

linshokaku:
When the trigger is given as a float in epoch units, there can be cases where it may not function correctly. Thus, we will make necessary adjustments to correct this.


emcastillo (approved):
Good catch!!

PR #740, opened 2023-08-15
`ppe.compile` Enable forward only and custom decompositions

emcastillo:
In some cases we want to generate a forward only graph using the aot lightweight tracing.
Also it may be necessary to avoid using default decompositions to prevent some operators to be decomposed and losing important information.


pytorch_pfn_extras/_dynamo/_splitter.py
```diff
@@ -131,3 +149,49 @@ def _no_partition(
...
+            filter(_is_primal, fwd_graph.graph.nodes)
+        )
+        for i, node in enumerate(primal_inputs):
+            node.name = (
+                self._parameter_names[i]
+                if i < len(self._parameter_names)
+                else f"input_{i - len(self._parameter_names)}"
+            )
+
+        # Remove the outputs
```
linshokaku, line 189: I want some detailed comments about the 'output_node' that is being deleted here. I believe it's an intermediate variable for gradient calculation, but is it sufficient to just delete one?

pytorch_pfn_extras/_dynamo/_splitter.py
```diff
@@ -131,3 +149,49 @@ def _no_partition(
...
+        output_node = [
+            node for node in fwd_graph.graph.nodes if node.op == "output"
+        ][0]
+        outputs = pytree.tree_flatten(output_node.args)[0]
+        fwd_graph.graph.erase_node(output_node)
```
linshokaku, line 194: It would be great if we could verify via tests whether or not the intended node has been deleted. However, if it proves difficult, I don't think there's a need to force its addition.

emcastillo: yeah! thats a great idea! thanks!!

emcastillo: added a check in the test!

pytorch_pfn_extras/_dynamo/_splitter.py
emcastillo: added comments explaining it!

pytorch_pfn_extras/_dynamo/_splitter.py
```diff
@@ -131,3 +149,70 @@ def _no_partition(
...
+        # Notice that the graph needs to return as many values as the inputs of the
+        # forward pass. The outputs of the joint graph returns additional values
+        # besides the gradients.
+        bwd_graph = torch.fx.Graph()
+        # Needs to create one gradient per each input element
+        bwd_outs = []
+        for i_node in primal_inputs:
+            bwd_outs.append(
+                bwd_graph.call_function(
+                    torch.ones, (i_node.meta.get("tensor_meta").shape,)
```
linshokaku, line 213: nit:
It seems more natural to output zeros.

linshokaku (approved):
LGTM

PR #742, opened 2023-08-23
Support chrome tracing in ppe.profiler

emcastillo:
Tracer class by @okdshin

The user should be the one in charge of enabling chrome for the records they want? I mean, passing a Emitter object instead of a boolean flag and avoiding to store in the object in TLS


emcastillo:
[trace.json.zip](https://github.com/pfnet/pytorch-pfn-extras/files/12450095/trace.json.zip)
You can check this file with Chrome and `about:tracing`

emcastillo:
Discussed offline

- How about merging the tracer with `TimeSummary`?
-  The TimeSummary class is complicated so we don't want to modify it more ..., also the TS object lives in the TLS too

kmaehashi:
Added me to assignee as I'd like to join the design discussion :)

emcastillo:
I changed the code to allow users to create their own tracers for timelines (not to rely only on google tracing format) and to allow users to supply a `tracer` object instead of using the default one in the TLS if needed

pytorch_pfn_extras/profiler/_tracing.py
```diff
@@ -0,0 +1,123 @@
...
+                duration_ns = time.perf_counter_ns() - begin_ns
+                self._event_list.append(
+                    dict(
+                        name=name,
+                        cat="",
+                        ph="X",
+                        ts=begin_ns / 1000,  # nano sec -> micro sec
+                        dur=duration_ns / 1000,  # ditto
+                        pid=0,
+                        tid=0,
```
linshokaku, line 72: Why isn't the pid and tid retrieved?

pytorch_pfn_extras/profiler/_tracing.py
```diff
@@ -0,0 +1,123 @@
...
+        pass
+
+    def flush(self, filename: str, writer: Writer) -> None:
+        pass
+
+
+class ChromeTracingSaveFunc:
+    def __call__(self, target: Dict[str, Any], file_o: Any) -> None:
+        log = json.dumps(target, indent=4)
+        file_o.write(bytes(log.encode("ascii")))
```
linshokaku, line 38: I believe it becomes bytes when encoded.

tests/pytorch_pfn_extras_tests/profiler_tests/test_record.py
```diff
@@ -127,3 +128,23 @@ def test_record_iterable_without_tag(device):
...
+@pytest.mark.parametrize("device", ["cpu", "cuda"])
+def test_record_iterable_with_chrome_trace(device):
+    if not torch.cuda.is_available() and device == "cuda":
+        pytest.skip()
+    model = torch.nn.Linear(30, 40)
+    model.to(device)
+
+    x = torch.arange(30, dtype=torch.float32).to(device)
+
+    x = torch.arange(30, dtype=torch.float32).to(device)
```
linshokaku, line 143: I think this can be deleted.

pytorch_pfn_extras/profiler/_record.py
```diff
@@ -43,13 +44,41 @@ def complete(self) -> None:
...
+    runtime_cls = runtime_registry.get_runtime_class_for_device_spec(device)
+    runtime_tracer = runtime_cls.trace
+
+    user_tracer: _tracing.Tracer
+    if isinstance(trace, bool) and not trace:
+        user_tracer = _tracing.DummyTracer()
+    elif isinstance(trace, bool):
+        user_tracer = _tracing.get_tracer()
+    elif isinstance(trace, _tracing.Tracer):
+        user_tracer = trace
```
linshokaku, line 68: If the trace object is customized by a user, this could potentially result in a non-thread-local object. Is this considered expected behavior? It might be beneficial to pass the Tracer class and then instantiate it via get_tracer.

pytorch_pfn_extras/profiler/_tracing.py
```diff
@@ -0,0 +1,123 @@
...
+
+_thread_local = threading.local()
+
+
+def get_tracer(tracer_cls: Type[Tracer] = ChromeTracer) -> Tracer:
+    if not hasattr(_thread_local, "_tracer"):
+        _thread_local._tracer = tracer_cls()
+    if _thread_local._tracer.__class__ is not tracer_cls:
+        raise TypeError("get_tracer called with a different cls")
+    return _thread_local._tracer  # type: ignore[no-any-return]
```
linshokaku, line 119: It would be beneficial to have a test that checks whether this implementation operates as expected during multithreading.

pytorch_pfn_extras/profiler/_record.py
emcastillo: Yeah, the user can have a tracer he/she manages explicitly like enabling or disabling it, or flush it when needed. So the final user should be in control of the object itself. I think it is ok if its not stored in thread local. We just used the TLS for dealing with the global object.  But maybe a single global variable is fine?

pytorch_pfn_extras/profiler/_tracing.py
emcastillo: By having the object in the thread local storage it is probably not going to work well, probably we should not have it in TLS and make the user instantiate the tracer explicitly ..., lets discuss it tomorrow!

linshokaku (approved):
LGTM!

kuenishi:
IMO, to integrate well with PFIO regarding https://github.com/pfnet/pfio/pull/326, it would be nicer to pass the `enable_trace` knob from trainer to underlying PFIO access IO connections such as `pfio.v2.S3` . You might not like TLS but it'd be nice for me...

pytorch_pfn_extras/training/extensions/timeline_trace.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def __init__(
+        self,
+        trigger: trigger_module.TriggerLike = (1, "epoch"),
+        filename: Optional[str] = None,
+        enable: Optional[trigger_module.TriggerLike] = None,
+        disable: Optional[trigger_module.TriggerLike] = None,
+        **kwargs: Any,
+    ):
+        self._tracer = kwargs.get("tracer", get_tracer())
```
linshokaku, line 60: I think it is more convenient for tracer to be included in the `__init__()` argument than for it to be kwargs, since it is available for type check.

```suggestion
    def __init__(
        self,
        trigger: trigger_module.TriggerLike = (1, "epoch"),
        filename: Optional[str] = None,
        enable: Optional[trigger_module.TriggerLike] = None,
        disable: Optional[trigger_module.TriggerLike] = None,
        tracer: Optional[Tracer] = None,
        **kwargs: Any,
    ):
        self._tracer = tracer if tracer is not None else get_tracer()
```

pytorch_pfn_extras/training/extensions/timeline_trace.py
```diff
@@ -0,0 +1,106 @@
...
+        self._disable = None
+        if disable is not None:
+            self._disable = trigger_module.get_trigger(disable)
+
+        self._trigger = trigger_module.get_trigger(trigger)
+
+        self._filename = "chrome_trace.json" if filename is None else filename
+        self._writer = kwargs.get("writer", None)
+
+    def _flush_trace(self, manager: ExtensionsManagerProtocol) -> None:
```
linshokaku, line 74: It would be nice to be able to select a mode to synchronize the tracer and then flush it as a strict `flush_trace()`. I would appreciate a TODO comment as there are so many things to consider including performance issues.

pytorch_pfn_extras/profiler/_tracing.py
```diff
@@ -0,0 +1,230 @@
...
+        raise NotImplementedError("Tracers must implement add_remote_event")
+
+    def clear(self) -> None:
+        raise NotImplementedError("Tracers must implement clear")
+
+    def flush(self, filename: str, writer: Writer) -> None:
+        raise NotImplementedError("Tracers must implement flush")
+
+    def enable(self, enable_flag: bool) -> None:
+        raise NotImplementedError("Tracers must implement enable")
```
linshokaku, line 28: The Tracer class should have a `finalize()` or `synchronize()` interface. In particular, if the QueueWorker is not synchronized, the possibility arises that the program will terminate before all profiles are output.

pytorch_pfn_extras/profiler/_tracing.py
```diff
@@ -0,0 +1,230 @@
...
+    def add_remote_event(
+        self, name: str, event: Dict[str, Union[str, int, float]]
+    ) -> None:
+        self._event_list.append(event)
+
+    def flush(self, filename: str, writer: Writer) -> None:
+        if not self._enable:
+            return
+        # TODO(ecastill): try to work on some append mode manipulating the
+        # file pointer and with json.dumps?
```
linshokaku, line 127: comment: 
I couldn't find any document, but it's worth checking to see if it supports the jsonl format.
The jsonl format is append friendly.

pytorch_pfn_extras/training/extensions/timeline_trace.py
```diff
@@ -0,0 +1,106 @@
...
+        state["_tracer"] = self._tracer.state_dict()
+        return state
+
+    def load_state_dict(self, to_load: Dict[str, Any]) -> None:
+        if hasattr(self._trigger, "load_state_dict"):
+            self._trigger.load_state_dict(to_load["_trigger"])
+        self._tracer.load_state_dict(to_load["_tracer"])
+
+    def finalize(self, manager: ExtensionsManagerProtocol) -> None:
+        self._flush_trace(manager)
```
linshokaku, line 103: I think that at the timing of finilize in extension, `synchronize()` in tracer should be executed.

pytorch_pfn_extras/profiler/_tracing.py
emcastillo: chrome does not support jsonl format for displaying the trace :(

pytorch_pfn_extras/profiler/_tracing.py
emcastillo: Added a finalize method, also enforced synchronization in `flush` method. Good catch!!!

emcastillo:
Thanks for the great review, PTAL!

linshokaku (approved):
LGTM

PR #752, opened 2023-10-03
Fix onnxruntime version for test

take-cheeze:
Ref: https://github.com/microsoft/onnxruntime/issues/17631


kmaehashi:
Would it be better to also make an issue for this one? to revert this PR later? 😄

take-cheeze:
Sounds nice not to forget it. 👍  Made an issue: https://github.com/pfnet/pytorch-pfn-extras/issues/753

PR #769, opened 2023-10-25
Fix `Logic.consume_options()` 

### LINKED ISSUE #767: Option argument is not applied during Logic initialization
It is possible to specify an `option` argument at the time of Logic class initialization.
https://github.com/pfnet/pytorch-pfn-extras/blob/3f023d33254a4744ea37fd5ea29724cfbb029d36/pytorch_pfn_extras/handler/_logic.py#L161

On the other hand, even if the options argument is not specified at the time of trainer initialization, it is treated as an empty dict.
https://github.com/pfnet/pytorch-pfn-extras/blob/3f023d33254a4744ea37fd5ea29724cfbb029d36/pytorch_pfn_extras/engine.py#L150

`Logic.cosume_options()` will update with default values for the empty dict values passed, so the options specified at initialization will no longer apply.
https://github.com/pfnet/pytorch-pfn-extras/blob/3f023d33254a4744ea37fd5ea29724cfbb029d36/pytorch_pfn_extras/handler/_logic.py#L185-L192

Consideration should be given to prohibiting `option` arguments at the time of initialization.

linshokaku:
close #767

### DIFF
--- a/pytorch_pfn_extras/handler/_logic.py
+++ b/pytorch_pfn_extras/handler/_logic.py
@@ -1,7 +1,17 @@
 import contextlib
 import dataclasses
 import warnings
-from typing import Any, Dict, Generator, Iterable, Mapping, Optional
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Generator,
+    Iterable,
+    List,
+    Mapping,
+    Optional,
+    Union,
+)
 
 import torch
 from pytorch_pfn_extras.handler._code_block import forward, update_parameters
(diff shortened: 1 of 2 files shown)


tests/pytorch_pfn_extras_tests/training_tests/test_trainer.py
```diff
@@ -878,7 +878,7 @@ def test_trainer_with_clousure_logic(device, progress_bar, path):
         evaluator=evaluator,
         extensions=extensions,
         out_dir=path,
-        logic=ppe.handler.ClousureLogic(options={"backward_outputs": ["loss"]}),
```
linshokaku, line 881: This test was not working properly in the master branch.

PR #776, opened 2023-10-30
Upload wheel

### LINKED ISSUE #770: Distribute wheel
Currently PPE is distributed as sdist, but it would be better to release it as whl as well.

linshokaku:
close #770

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Install Dependencies
         run: python3 -m pip install build --user
       - name: Build Packages
-        run: python3 -m build --sdist --outdir dist/ .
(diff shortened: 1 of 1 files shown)


kmaehashi:
Could you confirm that the contents of the wheel matches with the sdist?

linshokaku:
The following command was used to compare the generated package files, and nothing was output, confirming that there is no difference in content.

```sh
mkdir tar_package
tar -xzf pytorch-pfn-extras-0.7.3.tar.gz -C ./tar_package
unzip -q pytorch_pfn_extras-0.7.3-py3-none-any.whl -d ./whl_package
diff -r ./whl_package/pytorch_pfn_extras ./tar_package/pytorch-pfn-extras-0.7.3/pytorch_pfn_extras
```

PR #763, opened 2023-10-17
Running pytest in an MPI environment

linshokaku:
close #762 

depends on #765

PR #780, opened 2023-10-31
Fix on_error method in DistributedSnapshot

### LINKED ISSUE #779: `snapshot_on_error` in `DistributedSnapshot`
https://github.com/pfnet/pytorch-pfn-extras/blob/c5b4d58c16667d60e980ca96c5839416d63d50f2/pytorch_pfn_extras/training/extensions/_snapshot.py#L419-L428

If `snapshot_on_error` is specified in the snapshot extension, the `on_error()` method is called in case of an error, but in `DistributedSnapshot`, this method is not inherited and is not controlled by `saver_rank`.
This causes file corruption problems due to simultaneous writes, which needs to be corrected.

linshokaku:
close #779

### DIFF
--- a/pytorch_pfn_extras/training/extensions/_snapshot.py
+++ b/pytorch_pfn_extras/training/extensions/_snapshot.py
@@ -519,4 +519,4 @@ class _DistributedSnapshot(_Snapshot):
         if self._rank == self._saver_rank:
             super()._make_snapshot(manager)
         if self._size > 1:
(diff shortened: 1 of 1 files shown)


linshokaku:
Need to add test associated with MPI, so if #763 is merge first, add test.
If this one is merge first, add test to #763.

kmaehashi (approved):

PR #789, opened 2023-11-17
Trigger for using user custom functions

### LINKED ISSUE #774: Custom function trigger
Implement a helper Trigger class to be triggered by a user-implemented function.

```
class CustomTrigger(trigger.Trigger):
    class __init__(self, fn: Callable[..., bool], args: Sequential[Any], kwargs: Mapping[Any], check_trigger: TriggerLike) -> None
        ...
```

linshokaku:
close #774

### DIFF
--- a/pytorch_pfn_extras/training/triggers/function_trigger.py
+++ b/pytorch_pfn_extras/training/triggers/function_trigger.py
@@ -1,37 +1,44 @@
-from pytorch_pfn_extras.training._manager_protocol import ExtensionsManagerProtocol
-from typing import Callable, Optional, Sequence, Any, TYPE_CHECKING, Dict
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional, Sequence
+
 from pytorch_pfn_extras.training import trigger as trigger_module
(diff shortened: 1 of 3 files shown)


pytorch_pfn_extras/training/triggers/function_trigger.py
```diff
@@ -0,0 +1,54 @@
...
+        self._fn = fn
+        self._args = args or []
+        self._kwargs = kwargs or {}
+        self._interval_trigger = trigger_module.get_trigger(trigger)
+
+    def __call__(self, manager: ExtensionsManagerProtocol) -> bool:
+        if not self._interval_trigger(manager):
+            return False
+
+        return self._fn(*self._args, **self._kwargs)
```
emcastillo, line 37: I wonder if it would be good to add something like the current iteration/epoch as positional arguments

linshokaku: If iteration or epoch information is needed, it is assumed that users will implement a function that explicitly takes `manager` as an argument, and pass `manager` to `args`/`kwargs` to achieve this.

PR #790, opened 2023-11-20
[WIP] [ONNX] Add dynamo exporter support to export_testcase

take-cheeze:


take-cheeze:
Supporting interface of torch.onnx.dynamo_export is difficult now so closing

PR #812, opened 2024-03-21
Use argument `out_dir` not `self.out_dir`

nzw0301:
I'm not sure this is intended, but the currently we cannot specify `out_dir` from the argument because `self.out_dir` is always used.


linshokaku:
We have considered revisions in #735, however, either of the following is required:
- A change in the behavior that is non-obvious to users
- Breaking backwards compatibility
  - We believe there may be implications for users who utilize subclasses inheriting the Writer class.
 
Therefore, while the implementation can lead to misunderstandings, we concluded that it does not amount to a critical issue.

PR #817, opened 2024-05-08
Fix CuPy version installed in CI

kmaehashi:
Match the CuPy package for the current CUDA version. Unsure why it was working...


linshokaku (approved):
thanks!

PR #824, opened 2024-07-24
Avoid using `pkg_resources`

asi1024:
`pkg_resources` is deprecated. https://setuptools.pypa.io/en/latest/pkg_resources.html


linshokaku:
How about rewriting get_distribution().version to use import_lib.metadata.version() instead, which is available from Python 3.8+?

https://docs.python.org/ja/3/library/importlib.metadata.html

linshokaku (approved):
LGTM!

PR #834, opened 2024-10-15
added constraints to the ONNX version for Windows

linshokaku:
cf. https://github.com/onnx/onnx/issues/6267

Due to reports of the ONNX build for Windows being broken, we are temporarily limiting the version of ONNX.


kmaehashi:
Thanks, LGTM! Please proceed to merge after test pass.

(My guess is that `onnx<=1.16.1` should to be double-quoted, otherwise it will be treated as shell redirect 😉)

Issue #827, opened 2024-08-20
Enable export in external data format with ppe.onnx.


linshokaku:
In ppe.onnx, the destination for model output is specified by a file pointer when exporting onnx, but with torch>=1.11 onwards, along with the deprecation of the `use_external_data` flag, functionality has been added to output external data in the parent directory by specifying a file path when the model size exceeds a certain size. We will make modifications to be able to utilize this.

PR #839, opened 2024-10-16
Support for export of larger than 2GiB models

asi1024:
```
RuntimeError: The serialized model is larger than the 2GiB limit imposed by the protobuf library. Therefore the output file must be a file path, so that the ONNX external data can be written to the same directory. Please specify the output file name.
```


kmaehashi:
xref #827

PR #850, opened 2024-11-20
Fix Python version restriction

kmaehashi:
Fix python version restriction in `setup.py` to avoid uninstalling PPE v0.8.0 on unsupported Python environment.

Also bump version to v0.8.1 to cut a release and yank v0.8.0.


linshokaku (approved):
LGTM

PR #822, opened 2024-07-01
Add export_testcase to support length 1 arrays and lists

impactaky:
# Problem

The output allows a tuple or list of length 1, but not input.

# Reproduction

checkout 8defb94d093be26ccec2af3d4dc41d0c4c4283d0 and run
```
pytest tests/pytorch_pfn_extras_tests/onnx_tests/test_export_testcase.py -k test_export_tuple_input
```

# This PR

Add processing equivalent to output to input.

PR #852, opened 2024-12-03
[ONNX] Flatten input test data to support nested inputs

take-cheeze:
Original codes are from @impactaky -san. Thank you!

See also https://github.com/pfnet/pytorch-pfn-extras/pull/822


linshokaku (approved):
LGTM!

PR #855, opened 2025-03-26
Support PyTorch 2.6 & better compatibility with NumPy 2

kmaehashi:
* PyTorch 2.6 now defaults to `weights_only=True` when `torch.load`: https://pytorch.org/blog/pytorch2-6/
* PyTorch's Array interface (`__array__ `) is not yet compatible with NumPy 2: https://github.com/pytorch/pytorch/issues/136264


linshokaku (approved):
LGTM!

